PR #19, opened 2021-04-08
Features/hd wallet

iharob:


lib/src/solana_wallet.dart
```diff
@@ -1,35 +1,58 @@
...
+  Future<Signature> sign(List<int> data) async =>
       _ed25519.sign(data, keyPair: _keyPair);
 
   /// The address or public key of this wallet
+  static final _ed25519 = Ed25519();
+  final KeyPair _keyPair;
   final String address;
 
-  final KeyPair _keyPair;
+  static String _getHDPath(int account) => "m/44'/501'/${account - 1}'";
```
ookami-kb, line 45: I would rather place it into `HDKeyPair` class and let it accept `account`. It seems to be an implementation detail.

lib/src/types/hd_key_pair.dart
```diff
@@ -0,0 +1,40 @@
...
+  final List<int> _publicKey;
+
+  @override
+  Future<SimpleKeyPairData> extract() async => Future<SimpleKeyPairData>.value(
+        SimpleKeyPairData(
+          _private.key,
+          publicKey: await extractPublicKey(),
+          type: KeyPairType.ed25519,
+        ),
+      );
```
ookami-kb, line 30: No need to use `Future.value()` here, the value itself can be returned:

```suggestion  @override
  Future<SimpleKeyPairData> extract() async => SimpleKeyPairData(
        _private.key,
        publicKey: await extractPublicKey(),
        type: KeyPairType.ed25519,
      );
```

Same for the next methods here (just add async to the methods).

lib/src/solana_wallet.dart
```diff
@@ -1,35 +1,58 @@
...
+    this._keyPair, {
+    required this.address,
+  });
 
   /// Creates and initializes a new SolanaWallet for the given bip39
   /// [mnemonic] string of 12 words
-  static Future<SolanaWallet> fromMnemonic(String mnemonic) async {
+  static Future<SolanaWallet> fromMnemonic(
+    String mnemonic, {
+    int? account,
```
ookami-kb, line 17: Please add info about this parameter to the docstring. It's very confusing otherwise.

ookami-kb (changes requested):

lib/src/solana_wallet.dart
iharob: It is. I just didn't move it. Thanks for pointing it out!

lib/src/types/hd_key_pair.dart
iharob: It depends, without the async it is needed and that was the intention! because this function is not per-se async. Thanks for pointing it out!

lib/src/types/fees.g.dart
```diff
@@ -0,0 +1,15 @@
+// GENERATED CODE - DO NOT MODIFY BY HAND
```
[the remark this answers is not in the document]
iharob: Not sure how it got included :rofl:

lib/src/solana_wallet.dart
iharob: Agree!

lib/src/solana_wallet.dart
iharob: My previous comment was wrong. In fact. This is intended. The HD path is not part of the derivation or the key pair. It's in fact a parameter to the HD key pair object. Users of the library could want to use it for different tokens (in that case they would change the 501 in the path.

So basically. The derivation path is part of the wallet and not the key pair.

lib/src/types/hd_key_pair.dart
iharob: In this case there's an `await` so you are 100% correct. Thanks.

lib/src/solana_wallet.dart
ookami-kb: Yeah, makes sense.

lib/src/types/hd_key_pair.dart
ookami-kb: Any function returning `Future` is async, so next methods can be updated as well. E.g.:

```dart
Future<List<int>> extractPrivateKeyBytes() async => _private.key;
```

is just the shorter version of:

```dart
Future<List<int>> extractPrivateKeyBytes() =>
      Future<List<int>>.value(_private.key);
```

iharob: Actually the default crypto package had very similar code so I guess it's not so important.

ookami-kb: Definitely not important, just looks a little bit more readable to me.

ookami-kb (approved):

PR #23, opened 2021-05-04
feat!: Separate signing and sending transaction

iharob:
Now we can create transfer transactions and not send them immediately.


ookami-kb (changes requested):
Since `Message` and `SignedTx` are part of public code (they are used in wallet methods, they should be exported.

Also, I don't like method `compile` being part of `TransferTx`: all the classes in `types` directory are rather "output" types used by the client. Since we're creating "transfer" message, for me, it makes sense to make it a factory method of `Message` class:

```dart
factory Message.transfer({
  required Blockhash recentBlockhash,
  required int lamports,
  required String source,
  required String destination,
}) {
  final instruction = Instruction(
    programIdIndex: 2,
    accountIndices: CompactArray.fromList([0, 1]),
    data: CompactArray.fromList([
      ...2.toSolanaBytes(32),
      ...lamports.toSolanaBytes(64),
    ]),
  );

  final message = Message(
    header: MessageHeader(1, 0, 1),
    accounts: CompactArray.fromList([
      Address.from(source),
      Address.from(destination),
      Address.from(solanaSystemProgramID),
    ]),
    recentBlockhash: recentBlockhash.blockhash,
    instructions: CompactArray.fromList([instruction]),
  );
  return message;
}
```

In that case, clients can use it like this:

```dart
final message = Message.transfer(
  recentBlockhash: await solanaClient.getRecentBlockhash(),
  lamports: transferredAmount,
  source: sourceWallet.address,
  destination: targetWallet.address,
);
final SignedTx signedTx = await sourceWallet.signMessage(message);
```

Creating a method with "and" like "build**And**SignTransferTx" always looks like responsibility being spread. With separate methods it is more straightforward and re-usable: first, we create a message; second, we sign it.

lib/src/solana_serializable/signed_transaction.dart
```diff
@@ -1,17 +1,17 @@
...
+class SignedTx extends Serializable {
+  SignedTx({
     required this.signatures,
     required this.message,
   });
 
-  final CompactArray<Signature> signatures;
-  final Message message;
+  late final CompactArray<Signature> signatures;
+  late final Message message;
```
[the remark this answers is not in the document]
iharob: Makes total sense. Thanks for the observation. Also, late was needed before in a different implementation and just hangs out now. Good catch.

lib/src/solana_serializable/message.dart
```diff
@@ -13,6 +16,36 @@ class Message extends Serializable {
...
+      header: MessageHeader(1, 0, 1),
+      accounts: CompactArray.fromList([
+        Address.from(source),
+        Address.from(destination),
+        Address.from(solanaSystemProgramID),
+      ]),
+      recentBlockhash: recentBlockhash.blockhash,
+      instructions: CompactArray.fromList([instruction]),
+    );
+    return message;
```
ookami-kb, line 46:
```suggestion
    return Message(
      header: MessageHeader(1, 0, 1),
      accounts: CompactArray.fromList([
        Address.from(source),
        Address.from(destination),
        Address.from(solanaSystemProgramID),
      ]),
      recentBlockhash: recentBlockhash.blockhash,
      instructions: CompactArray.fromList([instruction]),
    );
```

lib/src/solana_wallet.dart
```diff
@@ -59,4 +57,37 @@ class SolanaWallet {
...
+
+    return SignedTx(
+      signatures: CompactArray.fromList([signature]),
+      message: message,
+    );
+  }
+
+  /// Create a transfer of [lamports] from this wallet to
+  /// [destination] transaction and sign it
+  Future<SignedTx> buildAndSignTransferTx({
```
ookami-kb, line 80: Let's remove this method. build**And**SignTransferTx is a code smell – a method should mainly do one thing.

Using it in client will be pretty straightforward:

```dart
final message = Message.transfer(
  recentBlockhash: await solanaClient.getRecentBlockhash(),
  lamports: transferredAmount,
  source: sourceWallet.address,
  destination: targetWallet.address,
);
final SignedTx signedTx = await sourceWallet.signMessage(message);
```

lib/src/util/solana_int_encoder.dart
```diff
@@ -0,0 +1,20 @@
+import 'package:convert/convert.dart';
+
+const int _bitsPerByte = 8;
+const int _hexCharsPerByte = 2;
+
+String _padTo(String what, String character, int length) =>
```
ookami-kb, line 6: This function is pretty useless – you can call `padLeft` directly.

lib/src/solana_client.dart
```diff
@@ -57,84 +55,31 @@ class SolanaClient {
...
-        s.Signature.from(await senderWallet.sign(serializedMessage));
-
-    final transaction = s.Transaction(
-      signatures: s.CompactArray.fromList([signature]),
-      message: message,
-    );
-
-    return _client.request(
-      simulate ? 'simulateTransaction' : 'sendTransaction',
+  /// Send signed transaction [signedTx]
```
ookami-kb, line 58: Docstring should describe what function does + dot at the end, so smth like:

```suggestion
  /// Sends signed transaction [signedTx].
```

lib/src/solana_client.dart
```diff
@@ -57,84 +55,31 @@ class SolanaClient {
...
-    SolanaWallet senderWallet,
-    String recipientAddress,
-    int amountInLamports,
-  ) async {
-    final data = await _doTransfer(
-      senderWallet,
-      recipientAddress,
-      amountInLamports,
-      simulate: true,
+  /// Simulates sending a signed transaction [signedTx]
```
ookami-kb, line 73:
```suggestion
  /// Simulates sending a signed transaction [signedTx].
```

ookami-kb (changes requested):

ookami-kb (approved):

PR #28, opened 2021-05-31
feat: Add commitment parameter to transactions methods

ookami-kb:


lib/src/solana_client.dart
```diff
@@ -207,11 +207,17 @@ class SolanaClient {
   ///
   /// Providing [before] and [until] also moves the cursor to a more specific
   /// subset.
+  ///
+  /// For [commitment] parameter description see
+  /// https://docs.solana.com/developing/clients/jsonrpc-api#configuring-state-commitment
+  ///
+  /// [TxStatus.processed] is not supported as [commitment].
```
iharobCP, line 214: What does this comment mean?

iharobCP (commented):
Can you take the

lib/src/solana_client.dart
ookami-kb: https://docs.solana.com/developing/clients/jsonrpc-api#getconfirmedtransaction

> (optional) Commitment; "processed" is not supported. If parameter not provided, the default is "finalized".

iharobCP: Ok, understood.

iharob (approved):

iharobCP (approved):

PR #29, opened 2021-06-09
fix: Forbid tansfers to from/to the same address

iharobCP:


antoineherzog:
Why should we forbid that? Technically it is possible

PR #34, opened 2021-06-17
feat: Derive associated token account address.

iharobCP:


lib/src/util/derive_associated_token_account_address.dart
```diff
@@ -0,0 +1,24 @@
+import 'package:cryptography/cryptography.dart' as crypto;
+import 'package:solana/src/base58/base58.dart' as base58;
+import 'package:solana/src/constants/constants.dart';
+
+const _nonce = 255;
+final _sha256 = crypto.Sha256();
+const _magicWord = 'ProgramDerivedAddress';
+
+Future<String> deriveAssociatedTokenAddress({
```
ookami-kb, line 9: Please add some documentation to it. It would be good to have a description of why it's needed at all.

lib/src/constants/constants.dart
```diff
@@ -0,0 +1,30 @@
+class SystemProgram {
```
ookami-kb, line 1: There shouldn't be instantiable classes that are not supposed to be instatiated.

It should be either top-level constants, like `systemProgramId` or at least abstract classes with private constructors:

```dart
abstract class SystemProgram {
  SystemProgram._();

  static const id = '11111111111111111111111111111111';
}
```

ookami-kb (changes requested):

lib/src/constants/constants.dart
iharobCP: Understood! Can you please just approve this and merge it? I will do it in my working tree as I had to merge this already with it so doing modifications to it can make things messy!

ookami-kb (approved):

PR #35, opened 2021-06-26
feat!: Add spl token support and an API to write arbitrary transactions 

iharobCP:


antoineherzog:
@iharobCP test are failing, can you fix?

antoineherzog:
and then ask the review of Kirill ?

lib/src/curve25519/edwards_point.dart
```diff
@@ -0,0 +1,55 @@
...
+    _FieldElement.one,
+    _FieldElement.one,
+    _FieldElement.zero,
+  );
+
+  late final _FieldElement _x;
+  late final _FieldElement _y;
+  late final _FieldElement _z;
+
+  // ignore: unused_field
```
ookami-kb, line 17: It's used.

lib/src/curve25519/field_element.dart
```diff
@@ -0,0 +1,1029 @@
...
+    _t[9] = h9;
+  }
+
+  const _FieldElement._fromConstList(this._t);
+
+  static const _FieldElement zero =
+      _FieldElement._fromConstList([0, 0, 0, 0, 0, 0, 0, 0, 0, 0]);
+  static const _FieldElement one =
+      _FieldElement._fromConstList([1, 0, 0, 0, 0, 0, 0, 0, 0, 0]);
+  static const _FieldElement minusOne = // ignore: unused_field
```
ookami-kb, line 95: If it's not used, why is it here?

lib/src/encoder/account_meta.dart
```diff
@@ -0,0 +1,107 @@
...
+          // Keep then one of the two that is either a signer or
+          // a writeable account
+          list[index] = item.mergeWith(list[index]);
+
+          return list;
+        }
+      });
+
+  /// Find an account with a matching [pubKey].
+  int indexOfPubKey(String pubKey) =>
```
ookami-kb, line 87: Extension like this on `Iterable` makes no sense, you need to have a list anyway. Let's rather make it an extension on `List<AccountMeta>` directly.

lib/src/encoder/compact_array.dart
```diff
@@ -0,0 +1,36 @@
+part of 'encoder.dart';
+
+class _CompactArray<T> extends ByteArray {
```
ookami-kb, line 3: It doesn't make sense for this class to be generic, it doesn't use this parameter inside.

lib/src/encoder/compact_array.dart
```diff
@@ -0,0 +1,36 @@
...
+  const _CompactArray(this._data);
+
+  factory _CompactArray._fromLengthAndContent(
+          ByteArray length, ByteArray content) =>
+      _CompactArray(Buffer.fromConcatenatedByteArrays([length, content]));
+
+  factory _CompactArray.fromIterable(Iterable<T> items) {
```
ookami-kb, line 10:
```suggestion
  factory _CompactArray.fromIterable(Iterable<Object> items) {
```

lib/src/encoder/compact_array.dart
```diff
@@ -0,0 +1,36 @@
...
+  final ByteArray _data;
+
+  @override
+  Iterator<int> get iterator => _data.iterator;
+}
+
+ByteArray _merge(ByteArray values, ByteArray next) =>
+    Buffer.fromConcatenatedByteArrays([values, next]);
+
+ByteArray _getBytesOf<T>(T value) {
```
ookami-kb, line 28:
```suggestion
ByteArray _getBytesOf(Object value) {
```

lib/src/encoder/compact_u16.dart
```diff
@@ -0,0 +1,24 @@
+part of 'encoder.dart';
+
+class _CompactU16 extends ByteArray {
+  _CompactU16(int value) {
+    final List<int> data = List<int>.empty(growable: true);
+    int rawValue = value;
+    // ignore: literal_only_boolean_expressions
```
ookami-kb, line 7: Useless `ignore`.

lib/src/encoder/extensions.dart
```diff
@@ -0,0 +1,59 @@
+part of 'encoder.dart';
+
+extension _AccountsIterableExt on Iterable<AccountMeta> {
```
ookami-kb, line 3: It doesn't make sense to name a private extension, it can just be `extension on Iterable<AccountMeta>`.

Same for other extensions.

lib/src/encoder/extensions.dart
```diff
@@ -0,0 +1,59 @@
...
+  /// to [Account Addresses Format][account addresses format], also removes
+  /// any duplicates.
+  ///
+  /// Duplicates are removed by picking signers and writeable accounts
+  /// over non-writeable and non-signers.
+  ///
+  /// Accounts are considered duplicates if they have the same public keys.
+  ///
+  /// [account addresses format]: https://docs.solana.com/developing/programming-model/transactions#account-addresses-format
+  Iterable<AccountMeta> _getAccounts(String? feePayer) =>
```
ookami-kb, line 53: If I got correctly what you mean here, I suggest changing this file to the following (and removing obsolete code from other files):

```dart
part of 'encoder.dart';

extension on Iterable<AccountMeta> {
  // Convert account metas to encoder public keys
  Iterable<Buffer> toSerializablePubKeys() =>
      map((a) => Buffer.fromBase58(a.pubKey));

  /// Removes all duplicate accounts, meaning accounts having the same public
  /// key. Signers and/or writeable accounts are picked over non-signers and
  /// non-writeable.
  List<AccountMeta> unique() =>
      fold<List<AccountMeta>>([], (List<AccountMeta> list, AccountMeta item) {
        final index = list.indexWhere((a) => a.pubKey == item.pubKey);
        if (index == -1) {
          list.add(item);
        } else {
          // Keep then one of the two that is either a signer or
          // a writeable account
          list[index] = item.mergeWith(list[index]);
        }

        return list;
      });
}

extension on Iterable<Instruction> {
  /// Prepares all the accounts from instructions:
  ///
  /// - extracts accounts from every instruction;
  /// - appends program id accounts (accounts with the sample public keys);
  /// - removes duplicates (by picking signers and writeable accounts
  ///   over non-writeable and non-signers);
  /// - sorts accounts according to [Account Addresses Format][1].
  ///
  /// [1]: https://docs.solana.com/developing/programming-model/transactions#account-addresses-format
  Iterable<AccountMeta> getAccounts(String? feePayer) {
    final accounts = expand((i) => [
          ...i.accounts,
          AccountMeta.readonly(pubKey: i.programId, isSigner: false),
        ]);

    return [
      ...accounts,
      if (feePayer != null)
        AccountMeta.writeable(pubKey: feePayer, isSigner: true),
    ].unique()
      ..sort(_accountComparator);
  }
}

extension on Blockhash {
  ByteArray toBytes() => Buffer.fromBase58(blockhash);
}

/// Compare accounts according to the following rules
///
/// Signer accounts go first, and within them writeable accounts
/// go first
///
/// Non-Signer accounts go after, and within them the writeable
/// accounts go first
int _accountComparator(AccountMeta a1, AccountMeta a2) {
  if (a1.isSigner && !a2.isSigner) return -1;
  if (!a1.isSigner && a2.isSigner) return 1;

  if (a1.isWriteable && !a2.isWriteable) return -1;
  if (!a1.isWriteable && a2.isWriteable) return 1;

  return 0;
}
```

I think it will make it more readable.

ookami-kb (changes requested):
Partial code review. Will continue later.

lib/src/encoder/compact_array.dart
iharobCP: True. It did use it initially. Well caught :)

lib/src/encoder/account_meta.dart
iharobCP: I want to avoid calling `toList()` so many times. When doing using like `where()` or similar methods that return `Iterable<T>` instead of `List<T>`. Will try to do it, but could be problematic.

lib/src/spl_token/associated_account.dart
```diff
@@ -0,0 +1,27 @@
+part of 'spl_token.dart';
```
ookami-kb, line 1: As a common notice: I don't like this organization with a lot of `part`/`part of` directives.

It breaks the idea of code separation and encapsulation since now all the classes are visible to each other and they can use private fields of each other. This is not really scalable. Also, it could be harder to control what to export from the library.

lib/src/utils.dart
```diff
@@ -0,0 +1,99 @@
+import 'package:cryptography/cryptography.dart';
+import 'package:solana/src/base58/base58.dart' as base58;
+import 'package:solana/src/curve25519/curve25519.dart';
+import 'package:solana/src/encoder/encoder.dart';
+
+typedef HashFunc = List<int> Function(List<int> m);
+
+abstract class Utils {
```
ookami-kb, line 8: Dart isn't Java, we don't need an abstract class to group static methods. These methods can be top-level functions.

ookami-kb (changes requested):

lib/src/spl_token/associated_account.dart
iharobCP: I would say it is exactly the opposite. I did this precisely to have library private things and export all the other things outside of the folder. I didn't know that private fields of classes are accessible to other classes in the same package, I wouldn't expect that to be the case. I think this simplified and put together the bits that belong together. Some classes are just helpers and are used as such in every "library" that was separated from the others.

lib/src/utils.dart
iharobCP: I wish it had namespaces so that we could have more semantic meaning in the places where we call them. Like identifying a package. I just wanted more organized code. I don't like the top level functions to be exported directly like that. But if it is the dart way, let's do it the dart way.

lib/src/spl_token/associated_account.dart
iharobCP: We should probably discuss this in a call as this is what made me more happy about the new organization. The fact that I don't need to export functions outside of the library is really a cool feature and I like to exploit it. It's like packages in go, which are also a great feature.

ookami-kb: Sure, let's discuss it in a call.

lib/src/utils.dart
ookami-kb: You can use named imports for that, e.g.

```dart
import 'smth' as utils;

utils.doSmth();
```

iharobCP: Yes, I am aware of it and in fact we are doing it inside the library with base58 encoder/decoder functions. I think it's more idiomatic so I will update it.

ookami-kb (approved):

PR #48, opened 2021-07-20
fix: Use the conventional $ to prefix generated function names

iharobCP:


ookami-kb:
@iharobCP tests are failing

PR #52, opened 2021-08-13
fix!: Remove explicit tokens adding

iharobCP:


packages/solana/lib/src/wallet.dart
```diff
@@ -243,39 +190,44 @@ class Wallet {
...
+    } on FormatException {
+      accounts = [];
+    }
+    if (accounts.isEmpty) {
       return false;
+    } else {
+      final indexOfExistingAccount =
+          accounts.indexWhere((a) => a.address == associatedTokenAddress);
+      return indexOfExistingAccount != -1;
     }
```
ookami-kb, line 219:
```suggestion
    return accounts.any((a) => a.address == associatedTokenAddress);
```

ookami-kb (changes requested):
Just a small suggestion from me.

But we also need to keep in mind that it's a breaking change.

ookami-kb (approved):

PR #53, opened 2021-08-16
feat!: Implement `getTokenAccountsByOwner` rpc method

iharobCP:


packages/solana/lib/src/rpc_client/signature_status.dart
```diff
@@ -33,11 +32,37 @@ class SignatureStatus {
...
+}
+
+@JsonSerializable(createToJson: false)
+class _SignatureStatusesResult {
+  _SignatureStatusesResult({required this.value});
+
+  factory _SignatureStatusesResult.fromJson(Map<String, dynamic> json) =>
+      _$SignatureStatusesResultFromJson(json);
+
+  @_NullableListConverter()
```
ookami-kb, line 51: What's so special about this list, why does need a custom converter?

packages/solana/lib/src/rpc_client/account.dart
```diff
@@ -69,11 +82,61 @@ class AccountData with _$AccountData {
...
+        default:
+          return AccountData.generic(data);
+      }
     } else {
       throw const FormatException('cannot decode account data');
     }
   }
+
+  @override
+  Map<String, dynamic> toJson(AccountData? object) => <String, dynamic>{};
```
ookami-kb, line 101: Since it's not actually intended for usage, it should rather throw `UnsupportedError`.

ookami-kb (changes requested):

packages/solana/lib/src/rpc_client/signature_status.dart
iharobCP: The generated `FromJson` for a `List<Something?>` does not take into account the fact that the template parameter is nullable. Leading to an exception about trying to convert `null` to `Something` (or to `Map<String, dynamic>`, I can't remember which exception is thrown).

iharobCP: This was the tricky thing I spoke about in the call.

packages/solana/lib/src/rpc_client/rpc_client.dart
```diff
@@ -320,7 +324,7 @@ class RPCClient {
   /// Returns Future that resolves to the statuses of a list of [signatures].
   /// Unless the [searchTransactionHistory] configuration parameter is included,
   /// this method only searches the recent status cache of signatures.
-  Future<Iterable<SignatureStatus?>> getSignatureStatuses(
+  Future<List<SignatureStatus?>> getSignatureStatuses(
```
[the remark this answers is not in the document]
iharobCP: I was thinking that providing it as a list could be more useful/normal for library consumers.

packages/solana/lib/src/rpc_client/rpc_client.dart
```diff
@@ -379,6 +384,38 @@ class RPCClient {
...
   }
 
+  /// Gets associated token accounts for a given user. If [mint] or [programId]
+  /// are provided, it returns the accounts associated to those specific values.
+  ///
+  /// For [commitment] parameter description [see this document][see this document]
+  /// [Commitment.processed] is not supported as [commitment].
+  ///
+  /// [see this document]: https://docs.solana.com/developing/clients/jsonrpc-api#configuring-state-commitment
+  Future<List<AssociatedTokenAccount>> getTokenAccountsByOwner({
```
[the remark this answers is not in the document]
iharobCP: Similar answer: I was thinking that providing it as a list could be more useful/normal for library consumers.

ookami-kb (approved):

Issue #56, opened 2021-08-20
Split the sign and send in the rpc client with an optional wait for status and not wait for status


iharobCP:
Create a convenience function for signing arbitrary messages and for sending previously signed transactions

PR #64, opened 2021-08-26
feat!: Create a separate sign method and split waiting for signature statuses from sending transactions

iharobCP:
Closes: #56 

Note: Also, some methods that are not part of the JSON RPC api were moved into an extension.


packages/solana/lib/src/rpc_client/rpc_client.dart
```diff
@@ -148,15 +149,16 @@ class RPCClient {
   ///
   /// [see this document]: https://docs.solana.com/developing/clients/jsonrpc-api#configuring-state-commitment
   Future<SimulateTxResult> simulateTransaction(
-    SignedTx signedTx, {
+    String transaction, {
     Commitment? commitment,
+    EncodingType encoding = EncodingType.base58,
```
ookami-kb, line 154: Why do we need to support `base58` encoding if it's deprecated?

ookami-kb (changes requested):

packages/solana/example/transfer_spl_token_example.dart
```diff
@@ -58,6 +58,7 @@ Future<void> example() async {
       source.signer,
     ],
   );
+  await rpcClient.waitForSignatureStatus(signature, TxStatus.finalized);
```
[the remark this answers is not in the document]
iharobCP: It does not anymore! that's part of what this PR does.

packages/solana/lib/src/rpc_client/rpc_client.dart
iharobCP: I see that eventhough deprecated it is still the default. Since this will now be THE library, I think we must support this. Perhaps mark it as deprecated would be the correct way.

packages/solana/lib/solana.dart
```diff
@@ -8,7 +8,7 @@ export 'src/spl_token/associated_account.dart';
 export 'src/spl_token/spl_token.dart';
 export 'src/system_program/system_program.dart';
 export 'src/token_program/token_program.dart';
-export 'src/utils.dart';
+export 'src/utilities.dart';
```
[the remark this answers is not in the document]
iharobCP: I do not remember why I did this, it was a few days ago and now I don't remember. I can rename it the way it was and if something makes me rename it again I will give an explanaiton, otherwise we can keep the original name.

packages/solana/example/transfer_spl_token_example.dart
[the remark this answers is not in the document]
iharobCP: It does look like I forgot to remove it. Thanks for pointing it out!

packages/solana/lib/src/rpc_client/rpc_client.dart
ookami-kb: Let's discuss it in discord then.

iharobCP: Updated!

ookami-kb (approved):

PR #67, opened 2021-08-30
fix: Remove commitment from `simulateTransaction` and `sendTransaction`.

iharobCP:


ookami-kb (changes requested):
- `simulateTransaction` actually accepts `commitment` parameter: https://docs.solana.com/developing/clients/jsonrpc-api#simulatetransaction
- `signAndSendTransaction` currently accepts `commitment` parameter but doesn't use it.
- `waitForSignatureStatus` doesn't accept `commitment` parameter, but it should be there.

packages/solana/lib/src/wallet.dart
```diff
@@ -40,7 +40,6 @@ class Wallet {
     final signature = await _rpcClient.signAndSendTransaction(
       message,
       [signer],
-      commitment: commitment,
     );
     await _rpcClient.waitForSignatureStatus(signature, TxStatus.finalized);
```
ookami-kb, line 44: We should rather use `commitment` parameter and fallback to `TxStatus.finalized` if only it's not set.

ookami-kb (changes requested):

antoineherzog (approved):
LGTM

ookami-kb (approved):

PR #78, opened 2021-10-03
feat: Export SignedTx

ookami-kb:
- Exported `SignedTx` and added `encodeBase58` method to `Signature` to be able to get the signature before actually sending the transaction.
- Turned `compile` into a method instead of a separate class, as the class does nothing useful.


packages/solana/lib/src/encoder/signature.dart
```diff
@@ -1,11 +1,14 @@
 import 'package:cryptography/cryptography.dart' as crypto;
+import 'package:solana/src/base58/encode.dart';
 import 'package:solana/src/common/byte_array.dart';
 
 class Signature extends ByteArray {
   Signature.from(crypto.Signature signature) : _data = signature.bytes;
 
   final ByteArray _data;
 
+  String encodeBase58() => base58encode(_data.toList(growable: false));
```
iharobCP, line 10: Lets call this `toBase58()` as do most of the libraries.

[web3](https://solana-labs.github.io/solana-web3.js/classes/PublicKey.html#toBase58) [Go](https://github.com/portto/solana-go-sdk#new-account--get-some-airdrop) [Java](https://github.com/p2p-org/solanaj/blob/f83f95b9b484b06495456d592659a1c90eb0448a/src/main/java/org/p2p/solanaj/core/PublicKey.java#L45).

iharobCP: I do know that this is not the `PublicKey` class but it is basically the same idea. We should also create a public key class eventually.

iharobCP (changes requested):

packages/solana/lib/src/encoder/signature.dart
ookami-kb: I was thinking about using the `Signature` class from `crypto` package directly instead. But it's a theme for discussion and for a different PR.

iharobCP (approved):

packages/solana/lib/src/encoder/signature.dart
iharobCP: Yes, probably a good call. We could add `toBase58()` as an extension only.

PR #118, opened 2021-11-24
feat: Complete rpc client implementation using a generator

iharobCP:


ChikinDeveloper:
Hi! If I'm right, with this implementation of Filter you have to filter by memcmp and dataSize, you can't filter by dataSize only (or memcmp only) in getProgramAccounts.

This is how I would implement it

```dart
abstract class Filter {
  Map<String, dynamic> toJson();
}

@JsonSerializable(createFactory: true, includeIfNull: false)
class DataSizeFilter implements Filter {
  const DataSizeFilter({required this.dataSize});

  factory DataSizeFilter.fromJson(Map<String, dynamic> json) =>
      _$DataSizeFilterFromJson(json);

  @override
  Map<String, dynamic> toJson() => _$DataSizeFilterToJson(this);

  /// Compares the program account data length with the provided
  /// data size.
  final int dataSize;
}

@JsonSerializable(createFactory: true, includeIfNull: false)
class MemcmpFilter implements Filter {
  const MemcmpFilter({
    required this.offset,
    required this.bytes,
  });

  factory MemcmpFilter.fromJson(Map<String, dynamic> json) =>
      _$MemcmpFilterFromJson(json);

  @override
  Map<String, dynamic> toJson() => _$MemcmpFilterToJson(this);

  /// Offset into program account data to start comparison.
  final int offset;

  /// Data to match, as base-58 encoded string and limited to
  /// less than 129 bytes.
  final String bytes;
}
```

Not really needed but you can also check that there is always at most 1 DataSizeFilter in a list, and check Memcmp filters conflicts.

packages/solana/lib/src/rpc/dto/account_data/account_data.dart
```diff
@@ -0,0 +1,75 @@
...
+      return ParsedAccountData.fromJson(data);
+    } else if (data is String) {
+      return AccountData._fromBase64String(data);
+    } else {
+      throw const FormatException('account data is in unknown format');
+    }
+  }
+
+  Map<String, dynamic> toJson() {
+    throw UnsupportedError('converting account data to json is not supported');
```
ookami-kb, line 73: If this is not supported, let's remove this method at all.

packages/solana/lib/src/rpc/dto/account_data/spl_token_program/mint_account_data_info.dart
```diff
@@ -0,0 +1,23 @@
+import 'package:json_annotation/json_annotation.dart';
+
+part 'mint_account_data_info.g.dart';
+
+@JsonSerializable()
```
ookami-kb, line 5: `toJson` is not used, should we remove it from generation? Same for other classes, I think for responses it doesn't make sense to keep it.

packages/solana/lib/src/rpc/dto/account_data/spl_token_program/token_program_account_data.dart
```diff
@@ -0,0 +1,34 @@
...
+    required String type,
+  }) = UnknownAccountData;
+
+  factory SplTokenProgramAccountData.fromJson(
+    Map<String, dynamic> data,
+  ) =>
+      _$SplTokenProgramAccountDataFromJson(data);
+
+  Map<String, dynamic> toJson() {
+    throw const FormatException('not serializable');
```
ookami-kb, line 32: Why? It's better to remove it at all.

packages/solana/lib/src/rpc/dto/account_data/empty_account_data.dart
```diff
@@ -0,0 +1,10 @@
+import 'package:solana/src/rpc/dto/account_data/account_data.dart';
+
+class EmptyAccountData implements AccountData {
```
ookami-kb, line 3: This is not used anywhere.

packages/solana/lib/src/rpc/dto/account_data/parsed_account_data.dart
```diff
@@ -0,0 +1,34 @@
...
+  const factory ParsedAccountData.unsupported(
+    Map<String, dynamic> parsed,
+  ) = UnsupportedProgramAccountData;
+
+  factory ParsedAccountData.fromJson(Map<String, dynamic> json) =>
+      _$ParsedAccountDataFromJson(json);
+
+  const ParsedAccountData._();
+
+  dynamic get info => map<dynamic>(
```
ookami-kb, line 29: I would remove this method at all, it's pretty confusing. If needed, one can do the manul mapping.

packages/solana/lib/src/rpc/dto/account_data/saved.g.dart
```diff
@@ -0,0 +1,24 @@
+// GENERATED CODE - DO NOT MODIFY BY HAND
```
ookami-kb, line 1: Is this file needed at all?

packages/solana/lib/src/rpc/dto/parsed_message/header.dart
```diff
@@ -0,0 +1,28 @@
...
+
+  factory Header.fromJson(dynamic json) {
+    if (json is List) {
+      return Header(
+        numRequiredSignatures: json[0] as int,
+        numReadonlyUnsignedAccounts: json[1] as int,
+        numReadonlySignedAccounts: json[2] as int,
+      );
+    } else {
+      return Header(
```
ookami-kb, line 17: You can combine manual and generated code:

```dart
@JsonSerializable(createToJson: false)
class Header {
  Header({
    required this.numRequiredSignatures,
    required this.numReadonlySignedAccounts,
    required this.numReadonlyUnsignedAccounts,
  });

  factory Header.fromJson(dynamic json) {
    if (json is List) {
      return Header(
        numRequiredSignatures: json[0] as int,
        numReadonlyUnsignedAccounts: json[1] as int,
        numReadonlySignedAccounts: json[2] as int,
      );
    } else {
      return _$HeaderFromJson(json as Map<String, dynamic>);
    }
  }

  final int numRequiredSignatures;
  final int numReadonlySignedAccounts;
  final int numReadonlyUnsignedAccounts;
}
```

packages/solana/lib/src/rpc/dto/block_commitment.dart
```diff
@@ -0,0 +1,25 @@
+import 'package:json_annotation/json_annotation.dart';
+
+part 'block_commitment.g.dart';
+
+/// The commitment of a bloc
+@JsonSerializable(createFactory: true, includeIfNull: false)
```
ookami-kb, line 6: `createFactory` is `true` by default.

ookami-kb (changes requested):
Review is still in progress, publishing intermediate comments.

packages/solana/lib/src/rpc/dto/account_data/account_data.dart
iharobCP: Good catch. It is here only because when using freezed it was required. But no longer because we are using `json_serializable`

packages/solana/lib/src/rpc/dto/account_data/spl_token_program/token_program_account_data.dart
iharobCP: Same reason. It was here for `@freezed`

packages/solana/lib/src/rpc/dto/account_data/empty_account_data.dart
iharobCP: That's a mistake. It should be.

packages/solana/lib/src/rpc/dto/account_data/parsed_account_data.dart
iharobCP: I agree.

packages/solana/lib/src/rpc/dto/account_data/saved.g.dart
iharobCP: Not at all. It was a temporary file. How did you see it? :fire:

packages/solana/lib/src/rpc/dto/parsed_message/header.dart
iharobCP: I love it ;)

packages/solana/lib/src/rpc/dto/account_data/empty_account_data.dart
iharobCP: It is used, in fact. I checked.

packages/jsonrpc_client/lib/src/client_generator.dart
```diff
@@ -0,0 +1,204 @@
...
+      final fromJsonParameters = [
+        '$data as Map<String, dynamic>',
+        if (genericFactory.isNotEmpty) genericFactory,
+      ];
+
+      if (typeIsPrimitive) {
+        return '$nullCheck$data as $typeName';
+      }
+
+      return '$nullCheck$typeName.fromJson(${fromJsonParameters.join(', ')})';
```
ookami-kb, line 149:
```suggestion
      final type = getDisplayString(withNullability: false);
      return '$nullCheck$type.fromJson(${fromJsonParameters.join(', ')})';
```

ookami-kb (changes requested):

packages/jsonrpc_client/lib/src/client_generator.dart
```diff
@@ -0,0 +1,204 @@
...
+  String _generateMethod(MethodElement method) {
+    final params = method.parameters.where((p) => p.isPositional).map((p) {
+      if (p.type.isNullableType) {
+        return 'if (${p.name} != null) ${p.toJson()}';
+      } else {
+        return p.toJson();
+      }
+    }).toList();
+    final isWithContext = method.metadata.any(
+      (a) => a.toSource() == "@withContext",
```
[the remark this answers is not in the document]
iharobCP: I didn't know this. Thanks.

analysis_options.yaml
```diff
@@ -1 +1,20 @@
 include: package:mews_pedantic/analysis_options.yaml
+
+dart_code_metrics:
```
ookami-kb, line 3: This can be removed as well.

ookami-kb (commented):
One comment, otherwise LGTM.

ookami-kb (approved):

PR #120, opened 2021-11-29
Fix nullable type handling

ChikinDeveloper:
Got error when calling getAccountInfo with an expected null result


ChikinDeveloper:
Fix nullable rpc client result type handling + fix filter implementation
We are running our tests on this branch, this will be the next release isn't it?
There's also an issue with rpc_client method conf serialization to json, we fixed it on our branch

packages/jsonrpc_client/lib/src/client_generator.dart
```diff
@@ -90,7 +92,7 @@ ${method.getDisplayString(withNullability: true)} async {
     );
   final dynamic value = $readerFn(response);
 
-  return ${returnType.fromJson("value")};
+  return ${(returnType.isNullableType) ? '(value == null) ? null : $returnTypeFromJsonStr' : returnTypeFromJsonStr};
```
iharobCP, line 95: Can we do this in the `fromJson()` method instead?

iharobCP (approved):
I had a small observation but it is not worh waiting for you to do it. I can do it before the other PR is reviewed.

PR #122, opened 2021-12-13
feat!: Use String feePayer instead of Keypair

ChikinDeveloper:
No need to know about the fee payer private key here


ChikinDeveloper:
warning • analysis_options.yaml:1:10 • The include file 'package:mews_pedantic/analysis_options.yaml' in \analysis_options.yaml' can't be found when analyzing 'ext_solana_dart_v2'. • include_file_not_found

But didn't change this file :/

antoineherzog:
@iharobCP can you check?

iharobCP:
Hello @ChikinDeveloper. Please pull master onto your branch so that checks can pass again. Then we can merge this PR. Thanks.

packages/solana/lib/src/helpers.dart
```diff
@@ -52,6 +52,16 @@ Future<String> findProgramAddress({
   throw const FormatException('cannot find program address with these seeds');
 }
 
+/// Create a program address for [programId] and [seeds]
+Future<String> createProgramAddress({
```
iharobCP, line 56: How is this different from `findProgramAddress()`?

iharobCP (changes requested):

packages/solana/lib/src/helpers.dart
Ltei: findProgramAddress will find a nonce and add it to the seeds. This method is needed when you already know the nonce and want to get a program address

ChikinDeveloper: findProgramAddress will find a nonce to append to the seeds. createProgramAddress is used when you already know the nonce.

ChikinDeveloper:
> Hello @ChikinDeveloper. Please pull master onto your branch so that checks can pass again. Then we can merge this PR. Thanks.

Looks like it's done, thanks antoine

packages/solana/lib/src/helpers.dart
iharobCP: I see. What is the use case for this? Can you perhaps remove this change from this PR and create a different PR? It is unrelated to the other change. We are very interested in merging this, so if you could that would be great. Thanks.

ookami-kb (approved):

PR #123, opened 2021-12-13
Add middlewares to JsonRpcClient

ChikinDeveloper:
Add middlewares to JsonRpcClient

We need to add our own rate limit manager to the client, and this is the least intrusive way to do it.
So we need at least a "onBeforeSend" method


iharobCP:
Hello @ChikinDeveloper. The team disucssed this PR and we consider this not to be a feature to have in the library. If you like you could try to do this before calling the methods by wrapping the rpc client with your own implementation that can count the number of requests.

ChikinDeveloper:
> Hello @ChikinDeveloper. The team disucssed this PR and we consider this not to be a feature to have in the library. If you like you could try to do this before calling the methods by wrapping the rpc client with your own implementation that can count the number of requests.

Hi ! 
Thats too bad, indeed we can to this but we have to override all the methods. 
Plus I believe it can be useful not only to add extra rate limit checks. You may want to add extra headers when using a custom rpc endpoint

PR #125, opened 2021-12-14
fix: Program account data filter serialization

iharobCP:


antoineherzog:
@ChikinDeveloper can you have a look?
it should fix https://github.com/cryptoplease/cryptoplease-dart/issues/124

ookami-kb (approved):

PR #129, opened 2021-12-18
Add createProgramAddress

ChikinDeveloper:
Add createProgramAddress helper. [(doc)](https://docs.rs/solana-program/1.4.4/solana_program/pubkey/struct.Pubkey.html#method.create_program_address)

Example use case : Get the vault signer key of a Serum dex
[The nonce is stored in the MarketState data](https://github.com/project-serum/serum-dex/blob/b96737fcda0585b18926e6f98277ebc33b8ca24f/dex/src/state.rs#L301)
[We then use createProgramAddress to find the address](https://github.com/project-serum/serum-dex/blob/b96737fcda0585b18926e6f98277ebc33b8ca24f/dex/src/state.rs#L1356)

### DIFF
--- a/packages/solana/lib/src/helpers.dart
+++ b/packages/solana/lib/src/helpers.dart
@@ -52,6 +52,17 @@ Future<String> findProgramAddress({
   throw const FormatException('cannot find program address with these seeds');
 }
 
+/// Create a program address for [programId] and [seeds]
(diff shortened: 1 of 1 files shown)


antoineherzog:
@ChikinDeveloper we already provide this function so we are closing the PR

PR #130, opened 2021-12-18
Use web_socket_channel package instead of dart:io for sockets

ChikinDeveloper:
dart:io can't be used on web. We are testing exclusively on web so we have an error ;)
We fixed it by using the [web_socket_channel  package](https://pub.dev/packages/web_socket_channel)

### DIFF
--- a/packages/solana/lib/src/subscription_client/subscription_client.dart
+++ b/packages/solana/lib/src/subscription_client/subscription_client.dart
@@ -1,6 +1,5 @@
 import 'dart:async';
 import 'dart:convert';
-import 'dart:io';
 
 import 'package:solana/src/exceptions/bad_state_exception.dart';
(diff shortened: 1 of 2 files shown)


packages/solana/lib/src/subscription_client/subscription_client.dart
```diff
@@ -38,6 +38,7 @@ class SubscriptionClient {
   int _lastRequestId = 1;
 
   void _dispatchMessage(dynamic data) {
+    print('_dispatchMessage : data=$data');
```
ookami-kb, line 41: Please remove `print`.

ookami-kb (changes requested):

antoineherzog:
@ChikinDeveloper Your change has been merged into #133

PR #131, opened 2021-12-22
fix!: Add missing imports and fix program accounts filter

iharobCP:


packages/solana/lib/src/rpc/extension.dart
```diff
@@ -1,6 +1,6 @@
 part of 'client.dart';
 
-extension Convenience on RpcClient {
+extension RpcClientConvenienceMethods on RpcClient {
```
ookami-kb, line 3: Let's just probably call it `RpcClientExt` – it seems to be a pretty common pattern.

packages/solana/lib/src/subscription_client/extension.dart
```diff
@@ -1,10 +1,15 @@
 part of 'subscription_client.dart';
 
-extension Convenience on SubscriptionClient {
+extension SubscriptionClientConvenienceMethods on SubscriptionClient {
```
ookami-kb, line 3: Same here, let's call it `SubscriptionClientExt`.

packages/solana/lib/src/rpc/extension.dart
iharobCP: Yes, of course!

ookami-kb (approved):

PR #133, opened 2021-12-24
refactor!: Update SubscriptionClient

ookami-kb:


packages/solana/lib/src/subscription_client/subscription_client.dart
```diff
@@ -17,155 +15,35 @@ import 'package:solana/src/subscription_client/notification_message.dart';
...
-  /// Connect to the websocket at [url] node.
-  static Future<SubscriptionClient> connect(String url) async =>
-      SubscriptionClient._(await WebSocket.connect(url));
-
-  /// Dispose this object and cancel any existing subscription.
-  void dispose() {
-    _subscription.cancel();
-  }
+  /// Subscribe to all incoming messages.
+  Stream<SubscriptionMessage> allMessages() => _stream.map(_parse);
```
iharob, line 33: I really don't see the use case. Normally the instantiator of this class will not care about handling messages manually. So this is probably not needed.

packages/solana/lib/src/subscription_client/subscription_client.dart
```diff
@@ -276,27 +156,121 @@ class SubscriptionClient {
...
+    final controller = StreamController<T>();
+
+    controller.onListen = () {
+      int? subscriptionId;
+      final requestId = _requestId++;
+
+      late final StreamSubscription subscription;
+
+      subscription = _stream.listen(
+        (event) {
```
iharob, line 209: Could we make this a method? I think it would be cleaner.

packages/solana/pubspec.yaml
```diff
@@ -16,6 +16,7 @@ dependencies:
   freezed_annotation: ^0.14.2
   http: ^0.13.3
   json_annotation: ^4.0.1
+  web_socket_channel: ^2.1.0
```
iharob, line 19: Isn't it better to merge the other PR first? So that the contribution of Chikin developer is not lost.

iharob (changes requested):

packages/solana/lib/src/subscription_client/subscription_client.dart
ookami-kb: I don't think it will make it cleaner – it uses scoped variables (`requestId` and `subscriptionId`).

packages/solana/pubspec.yaml
ookami-kb: This PR basically includes the changes from the other PR, so we can close the other one afterwards.

packages/solana/lib/src/subscription_client/subscription_client.dart
ookami-kb: There is a use case in the cryptoplease app. We need to have a single entrypoint, as the control over the connection should still be done by the client.

In case of using multiple subscriptions, we can simply do something like:

```dart
client.allMessages().listen(null, onDone: () {
  // connection is closed, reconnect
});
```

Also, it can be useful for centralized logging / error handling etc.

iharob: Yes, I like it. I was also wondering how we would do this. But this is precisely the way. Thanks.

iharobCP (approved):

PR #144, opened 2022-01-04
feat!: Add SolanaClient

ookami-kb:
Added `SolanaClient`, extracted token operations from `SplToken` to `SolanaClient`.

`SubscriptionClient` is created with each "waiting" operation, since it's needed for the lifetime of the method only.


packages/solana/lib/src/subscription_client/subscription_client.dart
```diff
@@ -32,6 +30,20 @@ class SubscriptionClient {
   /// Subscribe to all incoming messages.
   Stream<SubscriptionMessage> allMessages() => _stream.map(_parse);
 
+  /// Waits for transation with [signature] to reach [status].
```
iharobCP, line 33: Why not keep it as an extension? I'd like the classes to resemble their actual equivalents in the API.

packages/solana/test/solana_client_test.dart
```diff
@@ -49,7 +49,7 @@ void main() {
     );
     expect(signature, isNotNull);
     expect(
-      await solanaClient.getLamports(address: destination.address),
+      await solanaClient.rpcClient.getBalance(destination.address),
```
iharobCP, line 52: Can't we just create a `SolanaClient.getBalance()` method?

packages/solana/lib/src/subscription_client/subscription_client.dart
ookami-kb: Yeah, valid point. I'll revert it back to extension.

packages/solana/test/solana_client_test.dart
ookami-kb: Well, there's nothing special about this specific method, and if we'd go this way, we'll end up with proxying all the methods from `RpcClient` – I think it will be too much, having access to `rpcClient` seems to be simple enough.

iharobCP: Yes, probably right. I agree.

iharobCP (approved):

PR #157, opened 2022-01-14
feat: Add RPC client timeout

### LINKED ISSUE #156: we can add some configurable timeout there to be applied to all endpoints in RPC client.

ookami-kb:
Fixes #156

### DIFF
--- a/packages/solana/test/rpc_client_test.dart
+++ b/packages/solana/test/rpc_client_test.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:bip39/bip39.dart';
 import 'package:solana/solana.dart';
 import 'package:solana/src/rpc/dto/dto.dart';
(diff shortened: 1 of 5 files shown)


packages/solana/test/rpc_client_test.dart
```diff
@@ -9,6 +11,17 @@ import 'config.dart';
 const int _transferredAmount = 0x1000;
 
 void main() {
+  test('throws exception on timeout', () async {
+    final client = RpcClient(
+      devnetRpcUrl,
+      timeout: const Duration(milliseconds: 1),
```
iharobCP, line 17: Can you make it 0? Because I think we must make sure the test pass, and 1 ms despite being very little it is not zero.

iharobCP (changes requested):

packages/solana/test/rpc_client_test.dart
ookami-kb: Yeah, updated to `0`. Although, I would really like to see this test fail 😄

iharobCP: 1ms is a very short time indeed :laughing:

iharobCP (approved):

PR #167, opened 2022-01-28
feat: Add the ability to request all transactions in 1 request

iharobCP:


packages/solana/lib/src/rpc/extension.dart
```diff
@@ -32,18 +32,44 @@ extension RpcClientExt on RpcClient {
...
+  /// to the internal call to [RpcClient.getTransaction()]
+  Future<List<TransactionDetails>> getMultipleTransactions(
+    List<TransactionSignatureInformation> signatures, {
+    Commitment? commitment,
+    Encoding? encoding,
+  }) async {
+    final response = await _jsonRpcClient.bulkRequest(
+      'getTransaction',
+      signatures
+          .where((s) => s.err == null)
```
ookami-kb, line 57: Why do you skip transactions with errors?

packages/solana/lib/src/rpc/json_rpc_client.dart
```diff
@@ -15,50 +16,161 @@ class JsonRpcClient {
...
 
+    final response = await _postRequest(request);
+    if (response is _JsonRpcObjectResponse) {
+      return response.data;
+    } else {
+      throw FormatException('unexpected jsonrpc response type');
+    }
+  }
+
+  static FutureOr<Response> _onTimeout() {
```
ookami-kb, line 63: I don't see the profit of extracting this method.

packages/solana/lib/src/rpc/json_rpc_client.dart
```diff
@@ -2,6 +2,7 @@ import 'dart:async';
 import 'dart:convert';
 
 import 'package:http/http.dart' as http;
+import 'package:http/http.dart';
```
ookami-kb, line 5: There's already imported `http`.

ookami-kb (changes requested):

packages/solana/lib/src/rpc/json_rpc_client.dart
iharobCP: It should help to have the linter work this for us but great catch :D

packages/solana/lib/src/rpc/extension.dart
iharobCP: I havent thought this through honestly. Good catch.

packages/solana/lib/src/rpc/json_rpc_client.dart
iharobCP: BTW: it was just to combat the very ugly formatting that resulted otherwise

ookami-kb (approved):

Issue #175, opened 2022-02-08
Metadata.dart mintAccount and updateAuthority incorrect values


cryptobarbossa:
**Version**
0.18.0

**Issue**
The `getMetadata()` function returns 'updateAuthority' and 'mintAccount' with the incorrect corresponding values.

The 'mintAccount' value returned corresponds to 'updateAuthority' and similarly 'updateAuthority' instead returns the 'mintAccount' value.

iharob:
We'll surely take a look. Thanks.

iharob:
Ok, issue confirmed. We created this PR to fix it https://github.com/cryptoplease/cryptoplease-dart/pull/176

PR #176, opened 2022-02-08
fix: Fix the reading order for mintAccount and updateAuthority

iharob:
Closes: #175


ookami-kb:
Let's also cover it with tests.

PR #233, opened 2022-04-05
feat: add fee in dolar currency

rhbrunetto:
## Changes

- Fee is now being diplayed in dolar currency

Kazam_screenshot_00001
Kazam_screenshot_00000

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/cryptoplease/lib/presentation/conversion_rates.dart
```diff
@@ -25,6 +25,7 @@ extension ConversionRates on BuildContext {
     required FiatCurrency fiatCurrency,
     required Token token,
     required int amount,
+    int? fiatDecimals,
```
ookami-kb, line 28: Shouldn't be like that. We have `fiatCurrency`, so decimals should be taken from it. There should be no situation when these values are out of sync.

packages/cryptoplease/test/golden/screens/home_flow/send_flow/create_link_content_test.dart
```diff
@@ -11,6 +11,7 @@ void main() {
         amount: '1.25 SOL',
         fee: '0.01 SOL',
         fiatAmount: '123.45 USD',
+        fiatFee: '0.0001 USD',
```
ookami-kb, line 14: This looks wrong. `0.01` is the minimal possible amount for USD. For a smaller fee, it makes sense to format it as `< 0.01 USD`.

packages/cryptoplease/lib/presentation/screens/authenticated/send_flow/fungible_token/confirm_screen/components/fee_view.dart
```diff
@@ -0,0 +1,38 @@
+import 'package:cryptoplease/l10n/l10n.dart';
+import 'package:cryptoplease/presentation/screens/authenticated/send_flow/fungible_token/confirm_screen/components/list_item.dart';
+import 'package:cryptoplease_ui/cryptoplease_ui.dart';
+import 'package:flutter/material.dart';
+
+class FeeView extends StatelessWidget {
```
ookami-kb, line 6: Let's make this component smarter and more universal.

E.g. pass it `CryptoAmount` and `FiatCurrency` and it will render a list item with formatted amount and formatted converted currency. This way we can reuse it for both amount view and fee view (with just different styles).

ookami-kb (changes requested):

packages/cryptoplease/lib/presentation/screens/authenticated/send_flow/fungible_token/confirm_screen/components/fee_view.dart
rhbrunetto: Actually both fee and amount are `CryptoAmount`s once the fee itself is also in SOL. So, I created different constructors and let the logic of converting to fiat inside the widget.

packages/cryptoplease/test/golden/wrapper.dart
```diff
@@ -16,7 +16,7 @@ class Wrapper extends StatelessWidget {
             localizationsDelegates: AppLocalizations.localizationsDelegates,
             supportedLocales: AppLocalizations.supportedLocales,
             debugShowCheckedModeBanner: false,
-            home: child,
+            home: Center(child: child),
```
ookami-kb, line 19: Why is it needed? That looks weird to me.

packages/cryptoplease/lib/presentation/screens/authenticated/send_flow/fungible_token/confirm_screen/components/send_token_to_solana_address_content.dart
```diff
@@ -50,10 +45,11 @@ class SendTokenToSolanaAddressContent extends StatelessWidget {
               width: null,
             ),
             const SizedBox(height: 40),
-            ListItem(
+            AmountView.fee(
+              fee: fee,
               label: context.l10n.labelFee,
-              value: fee,
-            )
+              // formattedFiatFee: fiatFee,
```
ookami-kb, line 51: Please remove commented code.

ookami-kb (changes requested):

ookami-kb (approved):

PR #257, opened 2022-04-24
fix: Make Meta innerInstructions and logMessages nullable

ChikinDeveloper:
Ran into an issue while trying to fetch the following transaction with this library :
`4ZbnJUyNXTqSzN535p77fjsZEWC9fJWk3jBSLij8uuTh376mQjjviVUvuLFtFzDXGTsSHAjFAiViFAweLscrve8C`

Moreover, according to the [doc](https://docs.solana.com/developing/clients/jsonrpc-api#gettransaction), `innerInstructions` and `logMessages` are `<array|null>`

Edit: Not the same issue, but Buffer.fromuint64 is not supported in dart2js (web), and it looks like they aren't going to fix it. Unfortunately many utils like `TokenInstruction.transfer` will fail in web.
[https://github.com/dart-lang/sdk/issues/10275](https://github.com/dart-lang/sdk/issues/10275)

### DIFF
--- a/packages/solana/lib/src/rpc/dto/meta.dart
+++ b/packages/solana/lib/src/rpc/dto/meta.dart
@@ -35,7 +35,7 @@ class Meta {
 
   /// List of inner instructions or omitted if inner instruction
   /// recording was not yet enabled during this transaction.
-  final List<InnerInstruction> innerInstructions;
+  final List<InnerInstruction>? innerInstructions;
 
   /// List of token balances from before the transaction was
   /// processed or omitted if token balance recording was not yet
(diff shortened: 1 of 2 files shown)


ookami-kb:
> Edit: Not the same issue, but Buffer.fromuint64 is not supported in dart2js (web), and it looks like they aren't going to fix it. Unfortunately many utils like TokenInstruction.transfer will fail in web.

Yeah, for 64 we'll need to switch to `BigInt` and some custom flow. Ran into the issue with `borsh` serializer implementation as well.

ookami-kb (approved):

PR #276, opened 2022-05-22
feat: update `tokenAmount` and `splTokenTransferCheckedInfo` models

deimantasa:
## Changes

Current models have not included several fields, which seems to be existing in latest solana implementation.

Sample [transaction](https://explorer.solana.com/tx/3keW15m9Y5eHBrSrLdi8SoTD5AmPnGRku8Z4CRsCm3FL5KpcWjxtfhRoF6aiRxzndNRqrs3emoikd7V5AJQU8vjg?cluster=devnet).
Screenshot 2022-05-22 at 3 46 57 PM

After implementing changes, locally it works well.
Screenshot 2022-05-22 at 3 51 26 PM

To simulate this outcome, send [USDC](https://usdcfaucet.com/) to your `devnet` wallet. It seems that `checked transfer` has several more fields. However - I have not found any full official json for these models, therefore not sure, if they can be nullable or are fixed. Would be great if someone could guide me here.

Furthermore, after running solana test suite and running `flutter test` - it simply timed out. I guess I'm not doing something correctly. Would love to get some guidance here.

## Checklist

- [ ] Tests added.
- [x] Self-review done.

### DIFF
--- a/packages/solana/lib/src/rpc/dto/parsed_message/spl_token_transfer_checked_info.dart
+++ b/packages/solana/lib/src/rpc/dto/parsed_message/spl_token_transfer_checked_info.dart
@@ -12,6 +12,9 @@ class SplTokenTransferCheckedInfo {
     required this.tokenAmount,
     required this.source,
     required this.destination,
+    required this.mint,
+    required this.multisigAuthority,
+    required this.signers,
   });
 
   factory SplTokenTransferCheckedInfo.fromJson(
@@ -22,4 +25,7 @@ class SplTokenTransferCheckedInfo {
   final TokenAmount tokenAmount;
   final String source;
   final String destination;
+  final String mint;
+  final String multisigAuthority;
+  final List<String> signers;
 }
--- a/packages/solana/lib/src/rpc/dto/token_amount.dart
+++ b/packages/solana/lib/src/rpc/dto/token_amount.dart
@@ -9,6 +9,7 @@ class TokenAmount {
     required this.amount,
     required this.decimals,
     required this.uiAmountString,
+    required this.uiAmount,
   });
 
   factory TokenAmount.fromJson(Map<String, dynamic> json) =>
@@ -24,4 +25,7 @@ class TokenAmount {
 
   /// Token amount as a string, accounting for decimals.
   final String? uiAmountString;
+
+  /// Token amount as a decimal number.
+  final double uiAmount;
 }
--- a/packages/solana/lib/src/rpc/dto/parsed_message/spl_token_transfer_checked_info.g.dart
+++ b/packages/solana/lib/src/rpc/dto/parsed_message/spl_token_transfer_checked_info.g.dart
@@ -13,6 +13,10 @@ SplTokenTransferCheckedInfo _$SplTokenTransferCheckedInfoFromJson(
           TokenAmount.fromJson(json['tokenAmount'] as Map<String, dynamic>),
       source: json['source'] as String,
       destination: json['destination'] as String,
+      mint: json['mint'] as String,
+      multisigAuthority: json['multisigAuthority'] as String,
+      signers:
+          (json['signers'] as List<dynamic>).map((e) => e as String).toList(),
     );
 
 Map<String, dynamic> _$SplTokenTransferCheckedInfoToJson(
@@ -21,4 +25,7 @@ Map<String, dynamic> _$SplTokenTransferCheckedInfoToJson(
       'tokenAmount': instance.tokenAmount,
       'source': instance.source,
       'destination': instance.destination,
+      'mint': instance.mint,
+      'multisigAuthority': instance.multisigAuthority,
+      'signers': instance.signers,
     };
--- a/packages/solana/lib/src/rpc/dto/token_amount.g.dart
+++ b/packages/solana/lib/src/rpc/dto/token_amount.g.dart
@@ -10,6 +10,7 @@ TokenAmount _$TokenAmountFromJson(Map<String, dynamic> json) => TokenAmount(
       amount: json['amount'] as String,
       decimals: json['decimals'] as int,
       uiAmountString: json['uiAmountString'] as String?,
+      uiAmount: (json['uiAmount'] as num).toDouble(),
     );
 
 Map<String, dynamic> _$TokenAmountToJson(TokenAmount instance) =>
@@ -17,4 +18,5 @@ Map<String, dynamic> _$TokenAmountToJson(TokenAmount instance) =>
       'amount': instance.amount,
       'decimals': instance.decimals,
       'uiAmountString': instance.uiAmountString,
+      'uiAmount': instance.uiAmount,
     };


packages/solana/lib/src/rpc/dto/token_amount.dart
```diff
@@ -24,4 +25,7 @@ class TokenAmount {
 
   /// Token amount as a string, accounting for decimals.
   final String? uiAmountString;
+
+  /// Token amount as a decimal number.
+  final double uiAmount;
```
ookami-kb, line 30: AFAIK, `uiAmount` is deprecated, that's why we didn't include it.

packages/solana/lib/src/rpc/dto/parsed_message/spl_token_transfer_checked_info.dart
```diff
@@ -22,4 +25,7 @@ class SplTokenTransferCheckedInfo {
   final TokenAmount tokenAmount;
   final String source;
   final String destination;
+  final String mint;
+  final String multisigAuthority;
+  final List<String> signers;
```
ookami-kb, line 30: Not sure if this fields are nullable, will try to investigate it.

deimantasa: After some digging, this is what I understand:
1. `mint` is non-nullable
2. `multisigAuthority` is non-nullable
2. `signers` is nullable (maybe do empty list as a default)

Reference for `TransferChecked`.
https://docs.rs/solana-transaction-status/latest/src/solana_transaction_status/parse_token.rs.html#342

`mint` and `multisigAuthority` seems to be always included in the model.

However `signers` are included based on some condition. 
https://docs.rs/solana-transaction-status/latest/src/solana_transaction_status/parse_token.rs.html#477


Let me know if you see this same way?

packages/solana/lib/src/rpc/dto/token_amount.dart
deimantasa: Great spot. Thanks. Adding [official doc](https://docs.solana.com/developing/clients/jsonrpc-api#token-balances-structure) as clarification.

packages/solana/lib/src/rpc/dto/parsed_message/spl_token_transfer_checked_info.dart
ookami-kb: I agree that `mint` appears to be non-nullable.

But looking at `parse_signers` method, I think it provides **either** `multisig_field_name` (`"multisigAuthority"`) and `"signers"` **or** `owner_field_name` (`"authority"`).

packages/solana/lib/src/rpc/dto/token_amount.dart
ookami-kb: In that case, let's rather remove it from the DTO, unless there's some strong reason to keep it.

packages/solana/lib/src/rpc/dto/parsed_message/spl_token_transfer_checked_info.dart
deimantasa: Thanks for looking into it.

After closer inspection, I do agree with your conclusion. `SplTokenTransferInfo` model is also affected by it it seems.

I'll update code with changes in a bit.

packages/solana/lib/src/rpc/dto/parsed_message/spl_token_transfer_info.dart
```diff
@@ -19,4 +23,12 @@ class SplTokenTransferInfo {
...
   final String source;
   final String destination;
+  /// If [authority] is not available (null), then [multisigAuthority] and [signers] will be provided.
+  /// Otherwise [multisigAuthority] and [signers] will be [null] and [authority] will be provided.
+  final String? authority;
+  final String? multisigAuthority;
+  final List<String>? signers;
+  
+  /// If [authority] is not provided, [multisigAuthority] must be provided. And vice versa.
+  String get singleAuthority => authority ?? multisigAuthority!;
```
deimantasa, line 33: Added helper method to retrieve correct `authority`.

packages/solana/lib/src/rpc/dto/parsed_message/spl_token_transfer_checked_info.dart
```diff
@@ -22,4 +27,13 @@ class SplTokenTransferCheckedInfo {
...
   final String destination;
+  final String mint;
+  /// If [authority] is not available (null), then [multisigAuthority] and [signers] will be provided.
+  /// Otherwise [multisigAuthority] and [signers] will be [null] and [authority] will be provided.
+  final String? authority;
+  final String? multisigAuthority;
+  final List<String>? signers;
+
+  /// If [authority] is not provided, [multisigAuthority] must be provided. And vice versa.
+  String get singleAuthority => authority ?? multisigAuthority!;
```
deimantasa, line 38: Added helper method to retrieve correct `authority`.

packages/solana/lib/src/rpc/dto/token_amount.dart
deimantasa: Removed.

packages/solana/lib/src/rpc/dto/parsed_message/spl_token_transfer_checked_info.dart
deimantasa: Updated models + added references to Solana Rust models.

deimantasa:
Would be great if you could find some time (30min) for some 1-1 catchup - would love to properly spin test environment (as it does not seem to work for me now), because I'd like to also add tests for these new changes, as it's very important to not mess anything up.

ookami-kb:
Sure, find me in [Solana Dart developers](https://discord.gg/wK6WX7974J) Discord channel and let's schedule some call.

packages/solana/test/models_test.dart
```diff
@@ -0,0 +1,165 @@
...
+            source:
+                Ed25519HDPublicKey.fromBase58(senderTokenAccounts[0].pubkey),
+            mint: token.address,
+            destination: wallet.publicKey,
+            owner: source.publicKey,
+          ),
+        ),
+        signers: [source],
+        commitment: Commitment.finalized,
+      );
```
deimantasa, line 80: It keeps failing with 
```
jsonrpc-2.0 error (-32002): Transaction simulation failed: Attempt to debit an account but found no record of a prior credit.
	{accounts: null, err: AccountNotFound, logs: [], unitsConsumed: 0}
```

I assume I don't pass correct params or I'm missing something.

packages/solana/test/models_test.dart
```diff
@@ -0,0 +1,165 @@
...
+      amount: _tokenMintAmount,
+      mint: token.address,
+      authority: source,
+      commitment: Commitment.confirmed,
+    );
+  });
+
+  group('SplTokenTransferCheckedInfo test', () {
+    test('authority is not available', () {
+      // TODO
```
deimantasa, line 47: Not sure how to make `multisig` transaction. Because then `authority` will be `null` and `multisig` with `signatures` - won't. Would love to have helping hand here.

packages/solana/test/models_test.dart
```diff
@@ -0,0 +1,165 @@
...
+          as ParsedInstructionSplToken;
+      final parsedSplTokenInstruction = splTokenInstruction.parsed
+          as ParsedSplTokenTransferCheckedInstruction;
+      // TODO
+    });
+  });
+
+  group('SplTokenTransferInfo test', () {
+    test('authority is not available', () {
+      // TODO
```
deimantasa, line 106: Not sure how to make `multisig` transaction. Because then `authority` will be `null` and `multisig` with `signatures` - won't. Would love to have helping hand here.

deimantasa:
Added test commit with comments. Would love to hear suggestions on how to fix mentioned problems.

PR #285, opened 2022-06-04
feat: improve timeout message with more data provided

deimantasa:
## Changes

Adding more logging output in case rpc timeout happens.

Since RPC handled separately, it's hard to know which specific method failed (timed out). This addition includes more explicit logging in case RPC times out.

Before
```
[ERROR:flutter/lib/ui/ui_dart_state.cc(198)] Unhandled Exception: TimeoutException: request timed out
```

After
```
[ERROR:flutter/lib/ui/ui_dart_state.cc(198)] Unhandled Exception: TimeoutException: Time out. Request: {"jsonrpc":"2.0","id":"4","method":"getSignaturesForAddress","params":["GAmwwy4FzaHYzQTMLoYGtxAMumsV7u3RxrmrabtmmXaV",{"until":"67c9Q95bEJZgTQPTMsRwiqRSgdk3ezJoCbcLjX4pqGfhgegjo7TmNP53r1F1ttC8se6UwpqQg1xhdLEWrA63wiXG","commitment":"finalized"}]}
```

Now it's visible whole payload with exact method that timed out.

## Related issues

N/A

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/solana/lib/src/rpc/json_rpc_client.dart
```diff
@@ -71,9 +71,7 @@ class JsonRpcClient {
       body: body,
     ).timeout(
       _timeout,
-      onTimeout: () {
-        throw TimeoutException('request timed out');
-      },
+      onTimeout: () => throw TimeoutException('Time out. Request: $body'),
```
ookami-kb, line 74: I wouldn't put the whole body in exception message. For me, it makes more sense to create a custom exception class, e.g.:

```dart
class RpcTimeoutException extends TimeoutException {
  RpcTimeoutException(String url, this.body) : super('Request timed out: $url');

  final dynamic body;
}
```

and then just throw it with

```dart
throw RpcTimeoutException(_url, body);
```

ookami-kb (changes requested):

packages/solana/lib/src/rpc/json_rpc_client.dart
deimantasa: Updated.

ookami-kb:
@deimantasa I've updated the exception to save the body as a property instead of exposing it directly in the message.

Since exceptions are often automatically logged to some services, I would avoid passing the request body there implicitly. In case if it's desired behavior, it can always be done manually by capturing `RpcTimeoutException` and getting access to the body.

ookami-kb (approved):

deimantasa:
Hi @ookami-kb  - I do agree and disagree with this update.

Although it makes sense to outsource additional data capture, however, imagine scenario, that 20 RPC calls are being made and one of them timeout for some reason. User receives
```
[ERROR:flutter/lib/ui/ui_dart_state.cc(198)] Unhandled Exception: TimeoutException after 0:01:00.000000: RPC call timed out.
```

Then user needs to manually try catch each 20 RPC calls. Which I find troublesome, as there is not enough exposure, to what might have caused it.

I'd expect to see `at least` the name of the method within this exception. So basic message would be like
```
[ERROR:flutter/lib/ui/ui_dart_state.cc(198)] Unhandled Exception: TimeoutException after 0:01:00.000000: RPC call [accountBalance] timed out.
```

Then it's clear which call needs to be investigated further.

Let me know your thoughts.

ookami-kb:
@deimantasa there's probably nothing bad with exposing method name (apart from the fact the bulk requests need some special treatment), but I don't actually see that much sense in that in general. Can you please describe your use case in more detail?

Because, from my point of view, you shouldn't treat the timeout exception on a per request basis at all. This timeout doesn't indicate the problem with some specific request, it indicates the problem with a network connection. That means, if you need to know which exactly request timed out, you need to know the same about e.g. some `SocketException`.

> imagine scenario, that 20 RPC calls are being made and one of them timeout for some reason

Do you mean 20 RPC calls in different places or 20 RPC calls one after another? In the first case, you know which RPC call was that; in the second case, you don't need multiple catches, you catch it and process the details:

```dart
try {
  client.call1();
  client.call2();
  client.call3();
} on RpcTimeoutException catch (e) {
  print(e.body);
} on Exception catch (e) {
  // some other exception
}

deimantasa:
In my scenario, we have pretty robust but complex solana engine and we incorporate many various RPC calls for various reasons. We rely on various timers and subscriptions thus various RPC calls can get called at any given time.

I've received `TimeoutException` in one of them. And it was frankly, pain in the ass to understand which particular call times out. I had to manually wrap each RPC call to the `try catch` clause. And once I found the breaking piece - applied special treatment there.

If I had more visibility into this - it'd be immediately clear what's failing. `method` would help to immediately identify breaking piece.

ookami-kb:
Do you mean, it was some specific call that timed out always? That's weird, as I mentioned, it's just the transport layer problem, it shouldn't depend on the request itself 🤔 

Anyway, adding method doesn't seem like a hard task to add and won't do any harm I guess. I'll add this info.

deimantasa:
> Do you mean, it was some specific call that timed out always? That's weird, as I mentioned, it's just the transport layer problem, it shouldn't depend on the request itself 🤔
> 
> Anyway, adding method doesn't seem like a hard task to add and won't do any harm I guess. I'll add this info.

Yep, that's correct, one of our calls (`getSignaturesForAddress`) very frequently times out - and we've no idea why so far. Thus we needed to apply some different logic of how to go around this issue.

And yes - you're right, that's not the lib's issue - I'm just trying to point out that debugging to this particular problem would have been much easier with more exposed information. That'd have saved fair amount of time in our team for sure.


Thanks, @ookami-kb . Would you like me to take care of this follow up task, or you'll do it yourself?

ookami-kb:
I'll do it, should be an easyfix.

PR #329, opened 2022-07-17
feat: use roboto font as default

rhbrunetto:
## Changes

- Changes the app font to Roboto;
- Removes DIN font;
- Updates golden tests.

## Checklist

- [x] Tests added.
- [x] Self-review done.


packages/cryptoplease_ui/pubspec.yaml
```diff
@@ -4,31 +4,17 @@ version: 0.0.1
 publish_to: none
 
 environment:
-  sdk: ">=2.17.0 <3.0.0"
-  flutter: ">=3.0.0"
+  sdk: '>=2.17.0 <3.0.0'
+  flutter: '>=3.0.0'
```
ookami-kb, line 8: Should be changed.

packages/cryptoplease_ui/pubspec.yaml
```diff
@@ -4,31 +4,17 @@ version: 0.0.1
...
 
 dependencies:
   dfunc: ^0.7.2
   flutter:
     sdk: flutter
+  google_fonts: ^3.0.1
```
ookami-kb, line 14: It's better not to use this package, we don't need it to download the font over internet. Just use the same technique with fonts in assets.

ookami-kb (changes requested):

packages/cryptoplease_ui/lib/src/theme.dart
```diff
@@ -76,13 +76,12 @@ class CpThemeData {
         height: 1.25,
         color: primaryTextColor,
         fontFamily: 'Roboto',
-        package: 'cryptoplease_ui',
```
ookami-kb: Why was it removed? It won't work on iOS without it.

packages/cryptoplease_ui/lib/src/theme.dart
```diff
@@ -76,13 +76,12 @@ class CpThemeData {
...
       );
 
   ThemeData toMaterialTheme() => ThemeData(
         brightness: brightness,
         splashColor: CpColors.yellowColor.withOpacity(0.25),
-        fontFamily: 'packages/cryptoplease_ui/Roboto',
+        fontFamily: 'Roboto',
```
ookami-kb, line 84: Same here.

ookami-kb (changes requested):

ookami-kb (approved):

PR #333, opened 2022-07-18
replace exception with error in  findProgramAddress 

thomnico:
replace exception with error in  findProgramAddress  because it was stopping integration_test in debug mode when looking for a correct key. 
The debugger considered this exception "Uncaught" in vscode.

I did try this code with my usage and it works well.

Hope this helps,

## Related issues

Fixes #___

## Checklist

- [ ] PR title follows [Conventional Commits][1] guidelines.
- [ ] Tests added.
- [ ] Self-review done.

### DIFF
--- a/packages/solana/lib/src/crypto/ed25519_hd_public_key.dart
+++ b/packages/solana/lib/src/crypto/ed25519_hd_public_key.dart
@@ -5,6 +5,12 @@ import 'package:solana/src/encoder/byte_array.dart';
 import 'package:solana/src/encoder/encoder.dart';
 import 'package:solana/src/helpers.dart';
 
+class ProgramAddressResult {
+  ProgramAddressResult(this.oncurve, this.key);
+  final bool oncurve;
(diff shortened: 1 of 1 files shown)


ookami-kb:
Hey @thomnico 

For me, it doesn't make sense to return a result class here. First, it's a breaking change. Second, it's actually an error to provide invalid seeds, that's why the method throws an exception in that case.

In your case, I would rather advise intercepting an error in the test itself, if you consider some invalid input in test cases.

thomnico:
What you propose won't work at all as the function I am calling is findProgramAddress that is the one I was expected to fix to avoid the random uncaught exception during tests.

ookami-kb:
Can you please provide more details about your use case and/or some sample code? Because currently, I don't see why you can't use an additional try..catch block in your test code. Also, I don't really understand why you're providing invalid input data to `findProgramAddress` in your tests 🤷

PR #378, opened 2022-08-17
fix: goldens font rendering issue

### LINKED ISSUE #375: Golden tests don't render fonts properly
Investigate why golden tests don't render fonts after switching to `Roboto` and fix them.

moksh-mahajan:
## Changes

Goldens were not being rendered properly because, in the `cryptocase-ui` package, the value of the `fontFamily` property of the `ThemeData` as well as `_baseTextStyle` was not being set properly. These were pre-fixing the package name which is not required.

## Related issues

Fixes #375 

## Checklist

- [x] Self-review done.


ookami-kb:
@moksh-mahajan the problem with your solution is that it doesn't work on iOS, as you see the font is different:

<table>
<tr>
<td>

```dart
TextStyle get _baseTextStyle => TextStyle(
      height: 1.25,
      color: primaryTextColor,
      fontFamily: 'Roboto',
    );
```

</td>
<td>

```dart
TextStyle get _baseTextStyle => TextStyle(
      height: 1.25,
      color: primaryTextColor,
      fontFamily: 'Roboto',
      package: 'cryptoplease_ui',
    );
```

</td>
</tr>
<tr>
<td>

image

</td>
<td>

image

</td>
</tr>
</table>

moksh-mahajan:
@ookami-kb So the reason it was not working on iOS was: by default on iOS the default font family is `San Francisco` font but on Android, the default font family is already `Roboto`, that's why no changes were noticed on Android.

Now the issue is: apparently font generation in golden tests doesn't work with the `package` argument in the `TextStyle`. But if we remove the `package` argument the font doesn't work on the app.

So in order to make it work for both the app as well as golden test images, the solution I found was to declare all the fonts in the app's  `pubspec.yaml` referencing the `cryptoplease-ui` package.

Updated the PR as well.

ookami-kb:
I'd like to avoid duplicating font information in `app/pubspec.yaml`. I've found another way to do it in #379 – it just need a different font name, as `golden_toolkit` is doing some magic with font loading and when the name clashes with the default font (Roboto) it breaks the loading.

PR #441, opened 2022-09-18
feat: added outgoing split transfer status page

justinenerio:
## Changes

- Added Outgoing Split transfer status pages
- Added Timeline widget

## Screenshots/Videos
status

Canceled Flow

https://user-images.githubusercontent.com/28533130/192702615-607d051c-875e-4df8-9f83-762371ab7cca.mp4

Success Flow

https://user-images.githubusercontent.com/28533130/192702696-8f9e930a-7f67-4320-9a00-f31affee7be2.mov

## Checklist

- [x] PR is ready for review (if not, it should be a draft).
- [x] PR title follows [Conventional Commits][1] guidelines.
- [x] Screenshots/video added.
- [ ] Tests added.
- [x] Self-review done.

[1]: https://www.conventionalcommits.org/en/v1.0.0/

PR #451, opened 2022-09-22
feat: added CpTimeline widget

justinenerio:
## Changes

- Added CpTimeline widget 
- Added storybook example

## Related issues

[Fixes: #441 (comment)](https://github.com/cryptoplease/cryptoplease-dart/pull/441#discussion_r976875407)

## Screenshots
Screen Shot 2022-09-27 at 5 39 41 PM

Screen Shot 2022-09-27 at 5 39 51 PM

## Checklist

- [ ] Tests added.
- [x] Self-review done.


ookami-kb (changes requested):
There are some weird gaps between tiles:

image

Also, regarding the widget API – it should be more use-case oriented. Something like this:

```dart
class CpTimelineItem {
  CpTimelineItem({
    required this.title,
    required this.subtitle,
    this.trailing,
  });

  final String title;
  final String subtitle;
  final String? trailing;
}

enum CpTimelineStatus { inProgress, success, failure }

class CpTimeline extends StatelessWidget {
  const CpTimeline({
    super.key,
    required this.items,
    required this.status,
    required this.lastItem,
  });

  final List<CpTimelineItem> items;
  final CpTimelineStatus status;
  final int lastItem;

  @override
  Widget build(BuildContext context) {
    throw UnimplementedError();
  }
}
```

Everything else widget should be able to figure out on its own.

packages/cryptoplease_ui/lib/src/timeline_widget.dart
```diff
@@ -0,0 +1,183 @@
...
+  final String? trailing;
+}
+
+enum CpTimelineStatus { inProgress, success, failure }
+
+class CpTimeline extends StatelessWidget {
+  const CpTimeline({
+    super.key,
+    required this.items,
+    required this.status,
```
ookami-kb, line 22: What about `lastItem` parameter? Right now it's not possible to set which item is "active".

ookami-kb (changes requested):

justinenerio:
@ookami-kb i've fixed the weird gap between tiles. also reduced padding between lines, and added the white progress bar on in progress status

ookami-kb:
@justinenerio I've pushed some changes:

- Added `active` parameter. I remember we were discussing that it's not required for our specific use case, but it's easy to add and it feels more natural from the client's point of view.
- Extracted build methods into separate widgets.
- Added asserts for items list to be non-empty – I think it should be the client's responsibility to provide an empty widget.

ookami-kb (approved):

PR #617, opened 2022-11-14
feat: add top tokens by market cap in popular tokens section

justinenerio:
## Changes

- Updated popular token section to show top 20 coins by market cap
- Updated token details screen: token price taken from coingecko api rather than balancesbloc

## Video
https://user-images.githubusercontent.com/28533130/201679877-f3e9df58-a729-41fe-9444-c8a4b663b1c0.mov

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/cryptoplease/lib/features/popular_tokens/popular_token_list.dart
```diff
@@ -1,55 +1,87 @@
...
+                child: LoadingIndicator(),
+              ),
+            );
+
+            return state.when(
+              initial: () => loader,
+              processing: () => loader,
+              failure: (_) => SliverList(
+                delegate: SliverChildListDelegate(
+                  _defaultTokens.map((e) => _TokenItem(e, 0.0)).toList(),
```
ookami-kb, line 45: I don't think it makes sense to diplay stub tokens with `0` amount in case of a failure, it would make more sense to show the error instead.

Also, we really need to cache this info locally, but this can be done in another PR.

ookami-kb (changes requested):

ookami-kb:
@justinenerio please resolve conflicts

justinenerio:
@ookami-kb fixed merge conflicts

packages/cryptoplease/lib/features/popular_tokens/src/repository.dart
```diff
@@ -0,0 +1,48 @@
...
+          .toEither()
+          .mapAsync(
+            (responses) => IMap({
+              for (var e in responses) e.fromCoingecko(): e.currentPrice ?? 0
+            }),
+          );
+}
+
+extension on MarketsResponseDto {
+  Token fromCoingecko() => Token(
```
ookami-kb, line 38: @justinenerio 

This is enough for displaying token details (since this information comes from coingecko as well), but for buy/sell we need to know real token addresses, coming from `TokenList`.

It needs to be addressed before implementing buy/sell flow.

ookami-kb (approved):

PR #632, opened 2022-11-17
feat: generate swap transactions on backend to pay fees in usdc

### LINKED ISSUE #448: generate the jupiter swap instruction from our server (instead of the client wallet)

rhbrunetto:
## Changes

- Generates swap transaction on backend to pay fees in USDC;
- Backend retrieves swap fees.

## Related issues

Breakdown of #537 

Fixes #448 

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/cryptoplease_link/cryptoplease_api/pubspec.yaml
```diff
@@ -10,7 +10,8 @@ dependencies:
   dio: ^4.0.0
   freezed_annotation: ^2.0.1
   json_annotation: ^4.6.0
-  jupiter_aggregator: ^1.0.0
+  jupiter_aggregator:
+    path: jupiter_aggregator
```
ookami-kb, line 14: Please update it with the current master, I've moved the package.

packages/cryptoplease_link/lib/src/constants.dart
```diff
@@ -16,6 +16,7 @@ const moonpayHost = 'moonpay.$cpLinkDomain';
 const shareableLinkPaymentFee = 100000; // 0.1 USDC
 const directPaymentFee = 10000; // 0.01 USDC
 const directPaymentWithAccountCreationFee = 100000; // 0.1 USDC
+const swapFee = 100000; // 0.1 USDC
```
ookami-kb, line 19: `swapFee` is not a constant, no? I think the outcome of the discussion was that we have to do dynamic fees. Or maybe I'm confused it with something.

packages/cryptoplease_link/lib/src/payments/escrow_account.dart
```diff
@@ -1,3 +1,5 @@
+// ignore_for_file: deprecated_member_use
```
ookami-kb, line 1: What is deprecated?

packages/cryptoplease_link/cryptoplease_api/lib/src/client.dart
```diff
@@ -26,6 +26,11 @@ abstract class CryptopleaseClient {
     @Body() CreateDirectPaymentRequestDto request,
   );
 
+  @POST('/createSwapTransaction')
+  Future<SwapResponseDto> createSwapTransaction(
+    @Body() SwapRequestDto request,
```
ookami-kb, line 31: It shouldn't expose DTO from a separate package, create a separate DTO instead.

ookami-kb (changes requested):

packages/cryptoplease_link/lib/src/constants.dart
rhbrunetto: Should be fixed by https://github.com/cryptoplease/cryptoplease-dart/pull/638

packages/cryptoplease_link/lib/src/payments/escrow_account.dart
rhbrunetto: We need it since `CompiledMessage.fromSignedTransaction` is marked as deprecated in `solana: ^0.26.1`. Had to update it to use `SignedTx.decode` [here](https://github.com/cryptoplease/cryptoplease-dart/pull/632/files#diff-42c44042a8e0fb7a000a7bac97bf5247bdb974e99f61279d04efb00c83a316ea)

rhbrunetto:
Waiting for #641 to add price implementation.

packages/cryptoplease_link/lib/src/payments/escrow_account.dart
ookami-kb: You can use `SignedTx.fromBytes`. It doesn't make sense to deprecate this error since it's a package controlled by us.

packages/cryptoplease_link/lib/src/swap/create_swap_fee_payment.dart
```diff
@@ -0,0 +1,53 @@
...
+  required SolanaClient client,
+  required Ed25519HDPublicKey aReceiver,
+  required Ed25519HDPublicKey aSender,
+  required Commitment commitment,
+  required int amount,
+}) async {
+  // Swap fee is always paid in mainnet USDC
+  final mint = mainnetUsdc;
+
+  final shouldCreateAta = !await client.hasAssociatedTokenAccount(
```
ookami-kb, line 15: It's always platform, and it should always have ATA, so it doesn't make sense to check it here.

ookami-kb (changes requested):

ookami-kb (approved):

PR #633, opened 2022-11-18
feat: add Intercom support

### LINKED ISSUE #621: integrate intercom  support instead of contact by email

justinenerio:
## Changes

- Add Intercom support
- Updated minimum iOS target version to 13 

## Related issues

Fixes #621

## Video

https://user-images.githubusercontent.com/28533130/202617812-e422a7c0-9055-418a-b0ab-f01afe74506b.mp4

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/cryptoplease/lib/app/screens/authenticated/profile/components/danger_section.dart
```diff
@@ -13,6 +13,7 @@ class DangerSection extends StatelessWidget {
           sl<ISKPRepository>().clear();
           sl<ODPRepository>().clear();
           sl<OSKPRepository>().clear();
+          sl<IntercomService>().logout();
```
ookami-kb, line 16: What if the user will log out and then log in again? Intercom won't be re-initialized.

ookami-kb (changes requested):

packages/cryptoplease/lib/app/screens/authenticated/profile/components/danger_section.dart
justinenerio: You're right, I've updated pr to have a separate login call in onboarding after creation/restore of accounts

ookami-kb, line 16: This should rather be a deprecated way of having it here. Check `LogoutListener` and its usages instead.

ookami-kb (changes requested):

ookami-kb (approved):

PR #638, opened 2022-11-21
feat: calculate dynamic swap fees on backend

rhbrunetto:
## Changes

- Calculate swap fees in USDC

## Related issues

Breakdown of https://github.com/cryptoplease/cryptoplease-dart/pull/632

## Checklist

- [ ] Tests added.
- [x] Self-review done.

### DIFF
--- a/packages/cryptoplease_link/lib/src/swap/handler.dart
+++ b/packages/cryptoplease_link/lib/src/swap/handler.dart
@@ -7,13 +7,21 @@ import 'package:shelf/shelf.dart';
 import 'package:shelf_router/shelf_router.dart' as shelf_router;
 
 Handler addSwapHandler() =>
-    shelf_router.Router()..post('/getSwapFee', _handler);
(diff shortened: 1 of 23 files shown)


packages/cryptoplease_link/lib/src/swap/handler.dart
```diff
@@ -0,0 +1,27 @@
...
+
+Handler addSwapHandler() =>
+    shelf_router.Router()..post('/getSwapFee', _swapFeeHandler);
+
+Future<Response> _swapFeeHandler(Request request) async =>
+    processRequest<GetSwapFeeRequestDto, GetSwapFeeResponseDto>(
+      request,
+      GetSwapFeeRequestDto.fromJson,
+      (data) async {
+        final totalFeeInSol = data.route.fees?.totalFeeAndDeposits;
```
ookami-kb, line 17: What if the user will just provide "updated" route with lower fee?

rhbrunetto:
Merged into #632 since swap route generation and fee calculation will be done at the same call.

PR #685, opened 2022-12-26
fix: show low value decimal numbers

justinenerio:
## Changes

- Displays low value decimal numbers for Favorite List and Token Details 

## Screenshots
Simulator Screen Shot - iPhone 13 Pro - 2022-12-26 at 22 12 59

Simulator Screen Shot - iPhone 13 Pro - 2022-12-26 at 22 13 16

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/cryptoplease/lib/features/favorite_tokens/src/favorite_tokens_list.dart
```diff
@@ -145,7 +137,11 @@ class _TokenItem extends StatelessWidget {
         overflow: TextOverflow.ellipsis,
       ),
       trailing: Text(
-        tokenRate?.format(locale) ?? '-',
+        _formatPrice(
+          currentPrice?.toDouble(),
+          symbol: fiatCurrency.sign,
+          maxDecimals: 5,
```
ookami-kb, line 143: Why are there different max decimals for this list and for details screen?

And why is this logic for formatting repeated?

justinenerio: I've updated code to remove repeating logic and the same decimals.

Initially set them differently since list has smaller space compared to details, but upon rechecking it can be set the same.

ookami-kb (approved):

PR #739, opened 2023-01-24
feat: add balance refreshing after sending or receiving payments

### LINKED ISSUE #735: when we receive a payment with links or tips, pre-updated the homepage so the balance is already up to date (no need to pull to refresh))
when we receive a payment with links or tips, pre-updated the homepage so the balance is already up to date (no need to pull to refresh))

rhbrunetto:
## Changes

- Adds balance refreshing to OSKP, OT, ODP, IT, ISKP, PaymentRequest and Swap.

## Related issues

Fixes #735 

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/cryptoplease/lib/app/screens/authenticated/investments/investments_screen.dart
```diff
@@ -1,3 +1,5 @@
+import 'dart:async';
```
ookami-kb, line 1: Why is it needed?

packages/cryptoplease/lib/app/screens/authenticated/routes.dart
```diff
@@ -35,7 +35,10 @@ const authenticatedFlowRoutes = AutoRoute<dynamic>(
           page: EmptyRouterScreen,
           path: '',
           children: [
-            CustomRoute<void>(initial: true, page: InvestmentsScreen),
+            CustomRoute<void>(
+              initial: true,
+              page: InvestmentsScreen,
```
ookami-kb, line 40: Why?

packages/cryptoplease/lib/core/balances/bl/balances_bloc.dart
```diff
@@ -21,7 +21,7 @@ part 'balances_state.dart';
 
 final _logger = Logger('BalancesBloc');
 
-@injectable
+@lazySingleton
```
ookami-kb, line 24: BLoCs should never be singletons. In general, relying on service locator to ensure the scope of the entity should be avoided. We use SL only as a factory, scoping should be done by providers.

packages/cryptoplease/lib/features/incoming_split_key_payments/src/bl/iskp_bloc.dart
```diff
@@ -50,12 +53,16 @@ class ISKPBloc extends Bloc<_Event, _State> {
...
   final Ed25519HDKeyPair _account;
   final TxSender _txSender;
+  final BalancesBloc _balancesBloc;
 
   EventHandler<_Event, _State> get _handler => (event, emit) => event.map(
         create: (e) => _onCreate(e, emit),
         process: (e) => _onProcess(e, emit),
       );
 
+  void _refreshBalances() =>
```
ookami-kb, line 63: Let's rather avoid coupling `ISKPBloc` and `BalancesBloc`, the less BLoCs know about each other the better.

This glue functionality should rather be solved by a `BlocListener` of the `ISKPBloc` that sends the corresponding event to the `BalancesBloc`.

ookami-kb (changes requested):

ookami-kb (approved):

PR #787, opened 2023-02-11
feat: move encoder rpc client extensions

justinenerio:
## Changes

- add rpc_client_ext.dart to encoder exports

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/solana/lib/src/encoder/encoder.dart
```diff
@@ -12,5 +12,6 @@ export 'instruction.dart';
 export 'message/loaded_addresses.dart';
 export 'message/message.dart';
 export 'message_address_table_lookup.dart';
+export 'rpc_client_ext.dart';
```
ookami-kb, line 15: Let's rather move the code from here to `rpc/extension.dart`.

ookami-kb (changes requested):

PR #788, opened 2023-02-11
fix: remove black navigation bar

justinenerio:
## Changes

- removes bottom black navigation bar

## Checklist

- [ ] Tests added.
- [x] Self-review done.


ookami-kb:
This solution won't work – it puts the app into a full-screen mode (which trigger system notification) and the black bar re-appears when a user interacts with top/bottom bar. Check #790

PR #789, opened 2023-02-11
fix: update e2e tests

justinenerio:
## Changes

- update e2e test with new onboarding flow

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/espressocash_app/integration_test/app_test.dart
```diff
@@ -16,39 +14,20 @@ void main() {
...
   });
 }
 
 extension on WidgetTester {
-  Future<void> submit() async {
-    await tap(find.byKey(keyBottomButton));
-    await pumpAndSettle();
+  Future<void> pumpManual(int times) async {
+    for (int i = 0; i < times; i++) {
+      await pump(const Duration(seconds: 1));
```
ookami-kb, line 30: Why not just wait for the predefined duration instead of a loop?

justinenerio: It's because it takes more than one pump to finish the transition, and we cannot use `pumpAndSettle` because of the rive animation making it pump indefinitely.

ookami-kb (approved):

PR #840, opened 2023-02-23
fix: wrap tip link with dynamic links

justinenerio:
## Changes

- Add manual deep link url
- Wrap qr link and first link fallback with manual deep link

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/espressocash_app/lib/features/outgoing_split_key_payments/src/bl/tx_confirmed_watcher.dart
```diff
@@ -66,10 +66,12 @@ class _OSKPConfirmedJob extends CancelableJob<OutgoingSplitKeyPayment> {
 
     if (payment.amount.decimal <= _qrLinkThreshold) {
       final key = base58encode(privateKey.toList());
-      qrLink = SingleKeyPaymentData(
+      final rawLink = SingleKeyPaymentData(
         key: key,
         token: token.publicKey,
       ).toUri();
+
+      qrLink = await _linkShortener.manual(rawLink);
```
ookami-kb, line 74: Instead of `manual`, let's call this method `buildFullUrl`, and let's rename `shorten` to `buildShortUrl`.

Also, the OSKP first link building logic should be updated – instead of falling back to raw URL, it should fall back to full URL.

ookami-kb (changes requested):

PR #861, opened 2023-03-03
feat: support for custom http headers in RpcClient

EPNW-Eric:
It would be nice if we could add custom http headers to the RpcClient, and even override the default ones, so I added this very basic functionality.


packages/solana/lib/src/rpc/client.dart
```diff
@@ -17,8 +17,18 @@ abstract class RpcClient {
   factory RpcClient(
     String url, {
     Duration timeout = const Duration(seconds: 30),
+    Map<String, String> customHeaders = const {},
+    bool includeDefaultHeaders = true,
```
ookami-kb, line 21: Is there any use-case when the default header `Content-Type` **should not** be included?

EPNW-Eric: Yes, thats actually the reason for this PR. The default `Content-Type` header has `;charset=utf-8` included. I recently came across a solana RPC endpoint implementation that strictly requires `application\json` (see extrnode/extrnode#1). 

(Side note: Unfortunately, this PR will still not make me able to use extrnode, sine there is a bug in the dart http implementation itself, see dart-lang/http#184 and my PR dart-lang/http#886)

ookami-kb: I see, but you still need to include this header, you just need to alter it, right?

I suggest adding just `customHeader` parameter, and then using it like that in constructor:

```dart
headers: {..._defaultHeaders, ...customHeaders},
```

EPNW-Eric: I personally am fine with that. 

But just to clarify my intention: The idea behind that `bool` is to make it transparent to the users of the library that there might be other headers added by the library itself. I wanted to prevent a scenario where a user does not set a value in `customHeaders` and then starts wondering why outgoing requests contain something he has not added.

ookami-kb: In general, the library should not add any headers that are not absolutely necessary. `Content-Type` should be there anyway, but I would actually remove `charset` from the library. The official node seems to be working fine without this parameter.

And then, with the possibility to override the default header, you can still change it if needed.

EPNW-Eric: I've implemented your suggestions 😃

ookami-kb:
@EPNW-Eric can you please merge with the latest master branch? We've updated dependencies, so actions should run successfully now.

ookami-kb (approved):

PR #864, opened 2023-03-07
feat: cache coingecko results

justinenerio:
## Changes

Adds caching to coingecko api calls

## Caching Invalidation

Module | Path | Invalidate
-- | -- | --
Price | /simple/price | 1 min
Token Chart | /coins/{id}/market_chart | 1 hour
Token Details | /coins/{id} | 6 hours
Search | /search | 1 hour
  | /coins/markets |  

## Video
<details><summary>Video</summary>

https://user-images.githubusercontent.com/28533130/223729873-9d777836-46b9-4b38-8343-c8118ed163c7.mp4

</details>

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/espressocash_app/lib/features/conversion_rates/src/bl/token_price_cache.dart
```diff
@@ -0,0 +1,75 @@
...
+    for (final id in ids) {
+      final query = _db.select(_db.tokenPriceCacheRows)
+        ..where((p) => p.id.equals(id));
+
+      final row = await query.getSingleOrNull();
+
+      if (row != null) {
+        final now = DateTime.now();
+
+        if (now.difference(row.created) <= _maxAge) {
```
ookami-kb, line 32: Instead of retrieving and then comparing the age, you can add the condition to the query to select only records after specific DateTime. If nothing is found, just ignore it, it's not even necessary to remove the old record since it will be overwritten anyway.

packages/espressocash_app/lib/features/token_chart/src/data/token_chart_cache.dart
```diff
@@ -0,0 +1,67 @@
...
+
+class TokenChartCacheRows extends Table with EntityMixin {
+  TextColumn get data => text()();
+}
+
+extension on List<TokenChartItem> {
+  TokenChartCacheRow toDto(String id) => TokenChartCacheRow(
+        id: id,
+        created: DateTime.now(),
+        data: json.encode(map((e) => e.toJson()).toList()),
```
ookami-kb, line 55: Don't use json encoding, in case of migration it will fail with an error which will be hard to catch.

ookami-kb (changes requested):

packages/espressocash_app/lib/data/db/mixins.dart
```diff
@@ -15,6 +15,15 @@ mixin EntityMixin on Table {
   Set<Column<Object>>? get primaryKey => {id};
 }
 
+mixin QueryMixin on Table {
```
ookami-kb, line 18: Seems useless. You can use `EntityMixin` and just add `query` column directly.

packages/espressocash_app/lib/features/token_chart/src/token_chart_item.dart
```diff
@@ -10,4 +11,7 @@ class TokenChartItem with _$TokenChartItem {
   }) = _TokenChartItem;
 
   const TokenChartItem._();
+
+  factory TokenChartItem.fromJson(Map<String, dynamic> data) =>
```
ookami-kb, line 15: Is it still needed?

ookami-kb (changes requested):

packages/espressocash_app/lib/data/db/mixins.dart
justinenerio: Can't use `EntityMixin` for this case since `id` needs to be autoincrement integer.

justinenerio:
As discussed, we will use `dio_cache_interceptor` to handle caching.

Closing this PR, see https://github.com/espresso-cash/espresso-cash-public/pull/904

PR #893, opened 2023-03-24
feat: use smart contract for escrow payments backend

justinenerio:
## Changes

- adds smart contract escrow instructions
- adds new endpoints to use smart contract

## Checklist

- [x] Tests added.
- [x] Self-review done.


.github/workflows/check_pr.yml
```diff
@@ -62,6 +64,17 @@ jobs:
       - uses: actions/checkout@v2
       - uses: dart-lang/setup-dart@v1.3
 
+      - name: Install sudo package
+        run: apt update && apt install sudo
```
ookami-kb, line 68: Why is it needed?

.github/workflows/check_pr.yml
```diff
@@ -62,6 +64,17 @@ jobs:
...
+
+      - name: Start Solana Local Validator
+        uses: switchboard-xyz/solana-local-validator@v0.1
```
ookami-kb, line 71: Does it mean that we can move away from docker action?

.github/workflows/check_pr.yml
```diff
@@ -62,6 +64,17 @@ jobs:
...
+        with:
+          solana-version: stable
+          cluster: mainnet
+          args:
+            --bpf-program ${{ env.EC_LINKS_ID }} ${{ env.EC_LINKS_SO }}
```
ookami-kb, line 76: According to the action docs, you also need to run it, does it work without it?

packages/espressocash_backend/share_escrow/pubspec.yaml
```diff
@@ -0,0 +1,18 @@
+name: share_escrow
```
ookami-kb, line 1: Why does it need to be a separate package?

justinenerio: You're right, I though it was a good idea to separate the ix in another package. I've updated and moved share ix inside the main lib

.github/workflows/check_pr.yml
justinenerio: Yes the local validator should already be running after this action

.github/workflows/check_pr.yml
justinenerio: Yes since we cannot load solana programs using the docker action

.github/workflows/check_pr.yml
justinenerio: It is added since the solana-local-validator gh action uses sudo access

.github/workflows/check_pr.yml
justinenerio: Created https://github.com/espresso-cash/espresso-cash-public/pull/907 for this. Also apparently sudo installation is not needed.

ookami-kb (approved):

PR #894, opened 2023-03-24
feat: use smart contract for escrow payments app

justinenerio:
## Changes

- uses new endpoints that uses smart contract for payments

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/espressocash_app/lib/features/incoming_split_key_payments/src/bl/iskp_service.dart
```diff
@@ -69,7 +69,7 @@ class ISKPService {
         cluster: apiCluster,
       );
 
-      final response = await _client.receivePayment(dto);
+      final response = await _client.receivePaymentEc(dto);
```
ookami-kb, line 72: We should support both versions of links for some time.

ookami-kb (changes requested):

packages/espressocash_app/lib/features/incoming_split_key_payments/src/bl/iskp_service.dart
justinenerio: @ookami-kb updated code to support both. should have v3 links in app

ookami-kb (approved):

PR #904, opened 2023-03-27
feat: cache coingecko results

justinenerio:
## Changes

- Adds caching to coingecko api calls
- Uses `dio_cache_interceptor` package 

## Caching Invalidation

Module | Path | Invalidate
-- | -- | --
Price | /simple/price | 1 min
Token Chart | /coins/{id}/market_chart | 1 hour
Token Details | /coins/{id} | 6 hours
Search | /search | 1 hour
  | /coins/markets |  

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/espressocash_app/lib/core/coingecko_client.dart
```diff
@@ -0,0 +1,62 @@
...
+      DioCacheInterceptor(options: options),
+    );
+  }
+
+  Dio setMaxAge(Duration maxAge) => dio
+    ..interceptors
+        .insert(0, CacheInterceptor(options: options, maxAge: maxAge));
+}
+
+class CacheInterceptor extends Interceptor {
```
ookami-kb, line 34: Why not built-in `DioCacheInterceptor`?

justinenerio: It uses both `DioCacheInterceptor` and a custom `CacheInterceptor`. The library uses the response's cache headers to decide how to long cache. Since we do not have control over the server; and our use case would be to cache the data for custom x amount of time (ie cache details api for 6 hrs), a custom interceptor to handle max age is needed.

ref: https://github.com/llfbandit/dio_cache_interceptor/issues/6

packages/espressocash_app/lib/core/conversion_rates/bl/conversion_rates_client.dart
```diff
@@ -6,17 +6,22 @@ import 'package:freezed_annotation/freezed_annotation.dart';
...
+
 @injectable
 @RestApi(baseUrl: 'https://api.coingecko.com/api/v3/simple')
 abstract class ConversionRatesClient {
   @factoryMethod
-  factory ConversionRatesClient() =>
-      _ConversionRatesClient(Dio()..options.listFormat = ListFormat.csv);
+  factory ConversionRatesClient(CoingeckoClient client) =>
+      _ConversionRatesClient(
+        client.setMaxAge(_maxAge)..options.listFormat = ListFormat.csv,
```
ookami-kb, line 23: You're using the same singleton client, but overriding max age in several places. It should be done on per-request basis, e.g. with some `@Extra` annotation to the retrofit request.

ookami-kb (changes requested):

ookami-kb (approved):

PR #970, opened 2023-05-04
fix!: update solana seed derivation

### LINKED ISSUE #939: Key pair generated from seed phrase doesn't match to solana-keygen
There's no any way to produce the same key pair using solana-keygen prompt:// and Ed25519HDKeyPair.fromMnemonic

The reason is here

```dart 
Future<Ed25519HDKeyPair> fromMnemonic(
  String mnemonic,
  { int account = 0,
  int change = 0 }
)
```

As you can see, both account and change has default value of 0, while according to docs

> If account and change are both null, then we derive the address that would be returned by using
> 
> `solana-keygen pubkey prompt://`
> 

So, if we omit both account and change while calling fromMnemonic, we actually get the key derived from m/44'/501'/0/0 and it doesn't match to m/44'/501' that solana-keygen uses. Actually, we can't even force account = null as it is an int type value not int?.

Although this is a bug, I am not sure if it should be fixed as it will probably affect people using this library and break backward compatibility. I suggest documentation changes to describe clearly this peculiarity.

justinenerio:
## Changes

makes `account` and `change` nullable when creating seed using `fromMnemonic` to match behavior with solana cli

## Related issues

Fixes #939

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/solana/lib/src/crypto/ed25519_hd_keypair.dart
```diff
@@ -135,8 +135,18 @@ class Ed25519HDKeyPair extends KeyPair {
   }
 
   /// Build a derivation path with [account] and [change]
-  static String _getHDPath(int account, int change) =>
-      "m/44'/501'/$account'/$change'";
+  static String _getHDPath(int? account, int? change) {
```
rhbrunetto, line 138: I think we do a similar stuff on the app side [in this file](https://github.com/espresso-cash/espresso-cash-public/blob/0276db07afead3a34a98efbe8aa021f59f4efc5e/packages/espressocash_app/lib/core/wallet.dart#L30) regarding Saga wallets. Should we update there as well?

ookami-kb: That would be a breaking change for the app users, not sure if we should do it.

Also, in the app, we should actually provide `/0/0` when we call `fromMnemonic`, otherwise, again, it would be a breaking change for the app users.

justinenerio: Yes, for the app we have already set `/0/0` for local wallets

packages/solana/lib/src/crypto/ed25519_hd_keypair.dart
```diff
@@ -135,8 +135,18 @@ class Ed25519HDKeyPair extends KeyPair {
...
+    final path = StringBuffer("m/44'/501'");
+
+    if (account != null) {
+      path.write("/$account'");
+    }
+    if (change != null) {
+      path.write("/$change'");
+    }
```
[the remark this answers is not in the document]
justinenerio: I have updated the code. When `account` is null and `path` isn't, it writes `/0'`.

ookami-kb (approved):

PR #1022, opened 2023-07-04
feat: add ramp flow logic

justinenerio:
## Changes

- adds ramp flow on cash in/out to select a ramp provider based on country
- redirects to country selector if user country is null

## Video

<details>

https://github.com/espresso-cash/espresso-cash-public/assets/28533130/fc6cef95-ca26-490b-a2fa-6ab2f07e7645

</details>

## Checklist

- [ ] Tests added.
- [x] Self-review done.


antoineherzog:
@justinenerio can we just display the screen with the country the selector directly instead of the profile ? What do you think?

justinenerio:
@antoineherzog yea makes sense. will update

ookami-kb (approved):

PR #875, opened 2023-03-17
feat: recover pending escrow links

justinenerio:
## Changes

- recover pending escrow links even after wallet deletion

## Video

https://user-images.githubusercontent.com/28533130/225873094-1a3ceb00-1498-4765-bab3-078652612eb6.mp4

## Checklist

- [x] PR is ready for review (if not, it should be a draft).
- [x] PR title follows [Conventional Commits][1] guidelines.
- [x] Screenshots/video added.
- [ ] Tests added.
- [x] Self-review done.

[1]: https://www.conventionalcommits.org/en/v1.0.0/

PR #1104, opened 2023-10-23
feat: recover pending escrow link

justinenerio:
## Changes

recover pending escrow links even after wallet deletion

## Related issues

Update of https://github.com/espresso-cash/espresso-cash-public/pull/875 

## Checklist

- [ ] Tests added.
- [x] Self-review done.

### DIFF
--- a/packages/espressocash_app/lib/features/outgoing_link_payments/models/outgoing_link_payment.dart
+++ b/packages/espressocash_app/lib/features/outgoing_link_payments/models/outgoing_link_payment.dart
@@ -1,5 +1,6 @@
 import 'package:freezed_annotation/freezed_annotation.dart';
 import 'package:solana/encoder.dart';
+import 'package:solana/solana.dart';
 
 import '../../../core/amount.dart';
 import '../../../core/escrow_private_key.dart';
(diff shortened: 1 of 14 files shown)


packages/espressocash_app/lib/features/outgoing_link_payments/models/outgoing_link_payment.dart
```diff
@@ -90,4 +91,30 @@ class OLPStatus with _$OLPStatus {
...
   }) = OLPStatusCancelTxSent;
+
+  /// Tx was recovered from the blockchain.
+  /// These transactions cannot be resent, they can only be cancelled
+  const factory OLPStatus.recovered({
+    required EscrowPublicKey escrowPubKey,
+  }) = OLPStatusRecovered;
+
+  /// Similar to [OLPStatus.cancelTxCreated], but only contains public key
+  const factory OLPStatus.recoveredCancelTxCreated(
```
ookami-kb, line 102: Canceling a recovered payment shouldn't differ from canceling a normal payment. What about moving `EscrowPrivateKey` out of the `OLPStatus` into `OutgoingLinkPayment` and making it nulable? In that case, recovered payments will just differ in having a private key null.

packages/espressocash_app/lib/features/outgoing_link_payments/services/recover_pending_watcher.dart
```diff
@@ -0,0 +1,134 @@
...
+import '../../../../core/amount.dart';
+import '../../../../core/currency.dart';
+import '../../../../core/escrow_private_key.dart';
+import '../../activities/data/transaction_repository.dart';
+import '../../tokens/token.dart';
+import '../data/repository.dart';
+import '../models/outgoing_link_payment.dart';
+
+@injectable
+class RecoverPendingWatcher implements Disposable {
```
ookami-kb, line 20: I would say, this service should run just once on app start / after restoring wallet, doesn't make sense to listen to updates afterwards.

ookami-kb (changes requested):

PR #1114, opened 2023-10-30
feat: add experimental flag

justinenerio:
## Changes

- adds experimental flag based on list of addresses in remote config 
- list of address are be separted by comma

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/espressocash_app/lib/core/feature_flags.dart
```diff
@@ -18,4 +19,11 @@ class FeatureFlagsManager {
     );
     await _remoteConfig.fetchAndActivate();
   }
+
+  bool isExperimental({required String address}) => _remoteConfig
```
ookami-kb, line 23: The app shouldn't get the list of addresses directly. Instead, there's a parameter created in `RemoteConfig` console, and here you just get the bool value, so it will be just:

```dart
bool get isSomeFeatureEnabled => _remoteConfig.getBool('isSomeFeatureEnabled');
```

ookami-kb (changes requested):

ookami-kb (approved):

PR #1208, opened 2024-01-05
feat: add scalex withdraw endpoint

justinenerio:
## Changes

- adds endpoint for scalex withdraw
- update getFee endpoint to return withdrawal fee per partner

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/espressocash_api/lib/src/dto/withdraw.dart
```diff
@@ -0,0 +1,33 @@
...
+part 'withdraw.freezed.dart';
+part 'withdraw.g.dart';
+
+@freezed
+class WithdrawPaymentRequestDto with _$WithdrawPaymentRequestDto {
+  const factory WithdrawPaymentRequestDto({
+    required String senderAccount,
+    required String receiverAccount,
+    required int amount,
+    required int feePercentage,
```
ookami-kb, line 13: `feePercentage` shouldn't be set by this client, it's up for the backend to decide.

packages/espressocash_api/lib/src/dto/withdraw.dart
```diff
@@ -0,0 +1,33 @@
...
+import 'package:freezed_annotation/freezed_annotation.dart';
+
+part 'withdraw.freezed.dart';
+part 'withdraw.g.dart';
+
+@freezed
+class WithdrawPaymentRequestDto with _$WithdrawPaymentRequestDto {
+  const factory WithdrawPaymentRequestDto({
+    required String senderAccount,
+    required String receiverAccount,
```
ookami-kb, line 11: Let's rather have specific endpoints for now – this one for scalex. Different partners have different flows anyway, so we don't know for sure if this structure is universal. And for scalex, I think, we can just pass `partnerOrderId` and call the API on the backend to get all the details.

packages/espressocash_api/lib/src/dto/withdraw.dart
```diff
@@ -0,0 +1,33 @@
+import 'package:espressocash_api/espressocash_api.dart';
+import 'package:freezed_annotation/freezed_annotation.dart';
+
+part 'withdraw.freezed.dart';
+part 'withdraw.g.dart';
+
+@freezed
+class WithdrawPaymentRequestDto with _$WithdrawPaymentRequestDto {
+  const factory WithdrawPaymentRequestDto({
+    required String senderAccount,
```
ookami-kb, line 10: For new endpoints, `senderAccount` is not actually needed, API requests are signed by the wallet anyway.

ookami-kb (changes requested):

justinenerio:
@ookami-kb updated with all comments. should be a separate endpoint just for scalex

packages/espressocash_api/lib/src/dto/get_fees.dart
```diff
@@ -10,6 +10,7 @@ class GetFeesResponseDto with _$GetFeesResponseDto {
     required int escrowPayment,
     required int escrowPaymentAtaFee,
     required int splitKeyPayment,
+    required int withdrawalFeePercentage,
```
ookami-kb, line 13: This is a breaking change: until we deploy a new version of the backend, it will fail all the requests. Let's have it as `double withdrawalFeePercentage = 0` for now.

ookami-kb (changes requested):

PR #1221, opened 2024-01-11
fix: offramp watcher

justinenerio:
## Changes

- initialize watcher on offramp service
- add txId in offramp db for transaction matching

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/espressocash_app/lib/data/db/db.dart
```diff
@@ -259,6 +262,7 @@ class OffRampOrderRows extends Table with AmountMixin, EntityMixin {
   TextColumn get humanStatus => text()();
   TextColumn get machineStatus => text()();
   TextColumn get partnerOrderId => text()();
+  TextColumn get txId => text().nullable()();
```
ookami-kb, line 265: Why can't we take this information from `transaction`?

ookami-kb (approved):

PR #1378, opened 2024-04-19
feat: Tor/SOCKS5 proxy support: add optional proxyInfo property and use it (via the socks5_proxy package) if not null.

### LINKED ISSUE #1377: Tor/SOCKS5 proxy support for Dart Solana RpcClient
**Is your feature request related to a problem? Please describe.**
The Solana RpcClient should optionally support Tor/SOCKS5 proxies.

**Describe the solution you'd like**
An optional parameter accepting a proxy address (and port).

**Describe alternatives you've considered**
We could set up a local HttpClient (or similar) which would forward traffic to a Tor/SOCKS5 proxy and use that as our RpcClient node address, but not all platforms support local HttpClients equally (some need non-HTTPS options enabled, etc.)

**Additional context**
Stack Wallet wants to add Solana, but wants it to use Tor if that preference is enabled.

sneurlax:
Closes #1377.

Adds an optional proxyInfo property to a Solana RpcClient.  If it's null (as old clients will pass it by default), it changes no behavior.  If it's not null, it uses the provided proxyInfo to proxy requests.  Pass proxyInfo as a Map<String, dynamic> {String host, int port}.

Dart 3 supports Records, in which case proxyInfo could be `({String host, int port})? proxyInfo`.

## Changes

- Add the `socks5_proxy` package.  This is widely used and used by Stack Wallet.

## Related issues

Fixes #1377 

## Checklist

- [x] Screenshots/video not applicable.
- [x] Tests unaffected.
- [x] Self-review done.

### DIFF
--- a/packages/solana/lib/src/rpc/client.dart
+++ b/packages/solana/lib/src/rpc/client.dart
@@ -18,6 +18,7 @@ abstract class RpcClient {
     String url, {
     Duration timeout = const Duration(seconds: 30),
     Map<String, String> customHeaders = const {},
+    Map<String, dynamic>? proxyInfo, // {host: String, port: int}
   }) =>
       _RpcClient(
         url,
@@ -25,6 +26,7 @@ abstract class RpcClient {
           url,
           timeout: timeout,
           customHeaders: customHeaders,
+          proxyInfo: proxyInfo ?? {},
         ),
       );
 
--- a/packages/solana/lib/src/rpc/json_rpc_client.dart
+++ b/packages/solana/lib/src/rpc/json_rpc_client.dart
@@ -1,7 +1,9 @@
 import 'dart:async';
 import 'dart:convert';
+import 'dart:io';
+import 'dart:typed_data';
 
-import 'package:http/http.dart';
+import 'package:socks5_proxy/socks.dart';
 import 'package:solana/src/exceptions/http_exception.dart';
 import 'package:solana/src/exceptions/json_rpc_exception.dart';
 import 'package:solana/src/exceptions/rpc_timeout_exception.dart';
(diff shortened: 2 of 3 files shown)


ookami-kb (changes requested):
I don't think that adding another 3rd-party library for such a specific use case makes sense.

I would rather suggest making `JsonRpcClient._postRequest` method protected, and adding 

```dart
factory RpcClient.withCustomClient(JsonRpcClient client) {}
```

In that case, you can implement your custom behavior by extending `JsonRpcClient` class and passing it to the `RpcClient` factory.

sneurlax:
I am currently testing the changes @ookami-kb requested, should I just open a new PR @antoineherzog?

ookami-kb:
@sneurlax feel free to either reopen this one or create a new one - whatever is more convenient for you.

sneurlax:
There have been a series of issues implementing it as recommended.  The only solution I've got working as requested is very long because it has to implement overrides for every method in JsonRpcClient.  I really tried to minimize the diff and avoid these overrides, but the most elegant solution requires use of  `dart:mirrors` in order to get the expected return types of methods, which is not possible because out app is Dart 3, not the outdated Dart 2 which this package uses.  Dynamic return types and delegating methods in the custom RPC client using `noSuchMethod` might be a shorter solution, but I could not ultimately get that approach working in Dart 3.

I may be content to say "Stack Wallet is the only Dart/Flutter wallet to support Tor for Solana" rather than continue working it towards the requested changes.  We're simply using our cypherstack/espresso-cash-public fork to accomplish this currently using the code originally submitted as-is.  I don't like using an unmaintained fork, though, hence my attempt at fulfilling the requested code change.

@ookami-kb can you consider simply accepting the original patch?  socks5_proxy is one of the most widely used SOCKS packages and is Dart 2 and 3 compatible.  It would minimize the diff required for your Dart/Flutter users to have any option for proxy use.  Otherwise, the changes required to get it working are as shown on my subsequent PR (I can't reopen this one).

sneurlax:
update: as I have been testing more functions, I have been finding more issues.  I don't have enough Solana to test all functions and I really, really don't like this solution anyways.  See 1,000+ lines of biolerplate at: https://github.com/cypherstack/espresso-cash-public/blob/test/packages/solana/lib/src/rpc/custom_client.dart

@ookami-kb, what am I doing wrong?  You made this approach sound easy and simple, but it has been anything but easy and simple for me over here.

PR #1498, opened 2024-06-10
feat: store token list localdb

### LINKED ISSUE #1481: Store token list in local DB

Jaxiii:
## Changes

- TokenList now has local db storage
- The file token list is now CSV

## Related issues

Fixes #1481

## Checklist

- [x] Tests added.
- [x] Self-review done.


packages/espressocash_app/lib/features/tokens/data/token_dao.dart
```diff
@@ -0,0 +1,26 @@
+import 'package:drift/drift.dart';
+import '../../../data/db/db.dart';
+
+part 'token_dao.g.dart';
+
+@DriftAccessor(tables: [TokenRows])
+class TokenDao extends DatabaseAccessor<MyDatabase> with _$TokenDaoMixin {
```
ookami-kb, line 7: I don't see value in this class, this logic can be sitting directly in the repository.

packages/espressocash_app/lib/features/tokens/services/token_service.dart
```diff
@@ -0,0 +1,43 @@
...
+  Stream<List<TokenDTO>> watchAllTokens() => _tokenRepository.watchAllTokens();
+
+  Future<void> insertToken(TokenDTO tokenDTO) =>
+      _tokenRepository.insertToken(tokenDTO);
+
+  Future<void> updateToken(TokenDTO tokenDTO) =>
+      _tokenRepository.updateToken(tokenDTO);
+
+  Future<void> deleteToken(TokenDTO tokenDTO) =>
+      _tokenRepository.deleteToken(tokenDTO);
```
ookami-kb, line 24: These are just proxy methods to the repository, doesn't make sense to have a service for that.

packages/espressocash_app/lib/features/tokens/services/token_service.dart
```diff
@@ -0,0 +1,43 @@
...
+
+  Future<void> clearAllTokens() => _tokenRepository.clearAllTokens();
+
+  Future<void> initializeDatabaseWithJson(Map<String, dynamic> json) async {
```
ookami-kb, line 28: Where will this json come from? One of the main points is to get rid of having large json in memory.

It should rather be an assets csv file (versioned) that will be parsed on the app start whenever there's a new version.

ookami-kb (changes requested):

packages/espressocash_app/lib/features/tokens/token_list.dart
Jaxiii, line 1: Are this lines necessary? @ookami-kb
  
  ```// ignore: unused_field
  Iterable<TokenRow> _allTokensDB;
  
  Iterable<TokenRow> get tokensDB =>
      _allTokensDB.where((t) => t.chainId == chainId);```

packages/espressocash_app/assets/tokens/solana.tokenlist.csv
ookami-kb, line 1: Let's have titles as the first row, it will be easier to understand it.

packages/espressocash_app/lib/features/tokens/data/token_repository.dart
```diff
@@ -0,0 +1,26 @@
...
+
+@singleton
+class TokenListRepository {
+  const TokenListRepository(this._db);
+
+  final MyDatabase _db;
+
+  Future<List<TokenRow>> getAllTokens() => _db.select(_db.tokenRows).get();
+
+  Stream<List<TokenRow>> watchAllTokens() => _db.select(_db.tokenRows).watch();
```
ookami-kb, line 14: There shouldn't be methods like that, we don't want to load thousands of tokens into the memory.

packages/espressocash_app/lib/ui/user_avatar.dart
```diff
@@ -36,6 +36,7 @@ class CpUserAvatar extends StatelessWidget {
 }
 
 const TextStyle _textStyle = TextStyle(
+  fontFamily: 'RobotoApp',
```
ookami-kb, line 39: It should be default font style, no need to specify it explicitly.

packages/espressocash_app/lib/features/tokens/services/token_service.dart
```diff
@@ -0,0 +1,72 @@
...
+  final TokenListRepository tokenRepository;
+
+  Future<Either<Exception, void>> initializeDatabaseFromCsvFile(
+    String filePath,
+  ) =>
+      tryEitherAsync(
+        (_) async {
+          String? data = await rootBundle.loadString(filePath);
+
+          await tokenRepository.clearAllTokens();
```
ookami-kb, line 25: It should be done in a transaction and in a separate isolate.

packages/espressocash_app/lib/features/tokens/token_list.dart
```diff
@@ -34,23 +44,53 @@ class TokenList {
...
+    await service.initializeDatabaseFromCsvFile(
+      Assets.tokens.solanaTokenlist,
+    );
+
+    _allTokensDB = await service.tokenRepository.getAllTokens();
+  }
+
+  Future<void> initialize() async {
+    final tokenList = await service.tokenRepository.getAllTokens();
+    if (tokenList.isEmpty) {
```
ookami-kb, line 88: It shouldn't only be done once, but every time we update the file with tokens.

ookami-kb (changes requested):

packages/espressocash_app/lib/features/tokens/token_list.dart
```diff
@@ -1,58 +1,105 @@
...
-  final ParsedContent _parsedContent;
+  static final TokenListRepository _repository =
+      TokenListRepository(sl<MyDatabase>());
+
+  final TokenService service;
+
   final int chainId;
 
-  Iterable<Token> get _allTokens => _parsedContent.tokens;
+  Future<Token?> getTokenByMint(String mint) async {
```
ookami-kb, line 45: It never returns `null`.

Also, this method and `initialize` should rather be defined on repository, and all the clients should deal with repository directly.

packages/espressocash_app/lib/features/tokens/services/token_service.dart
```diff
@@ -0,0 +1,72 @@
...
+  final TokenListRepository tokenRepository;
+
+  Future<Either<Exception, void>> initializeDatabaseFromCsvFile(
+    String filePath,
+  ) =>
+      tryEitherAsync((_) async {
+        final String data = await rootBundle.loadString(filePath);
+        final List<TokenRow> tokenIterable =
+            await compute(_parseAndInsertTokens, data);
+        await tokenRepository.insertTokens(tokenIterable);
```
ookami-kb, line 25: You're still loading the whole (potentially) huge file on the main isolate, push the data to another isolate and then do the saving again on the main isolate. Instead, the whole operation should be done in the isolate, including inserting data into db. Also, preferably, the file itself should be parsed in chunks, not loaded into memory at once.

I would also put thin method directly into repository initialization.

ookami-kb (changes requested):

packages/espressocash_app/lib/features/tokens/token_list.dart
Jaxiii: Fixed in [4c03680](https://github.com/espresso-cash/espresso-cash-public/pull/1498/commits/4c03680f81d8b86eb2dc88cfbdf1da12b7d2940d)

packages/espressocash_app/lib/features/payment_request/widgets/request_success.dart
```diff
@@ -48,14 +40,19 @@ class RequestSuccess extends StatelessWidget {
...
-            CpTimeline(
-              status: CpTimelineStatus.success,
-              items: [
-                requestCreated,
-                moneyReceived,
-              ],
-              active: 1,
-              animated: false,
+            FutureBuilder(
+              future: request.payRequest.cryptoAmount(
```
ookami-kb, line 44: Never create future in `build` method – in the worst case it can be create with every frame.

packages/espressocash_app/lib/features/tokens/services/token_service.dart
```diff
@@ -0,0 +1,10 @@
+import 'package:injectable/injectable.dart';
+
+import '../data/token_repository.dart';
+
+@lazySingleton
+class TokenService {
```
ookami-kb, line 6: I guess it can be deleted?

packages/espressocash_api/lib/src/dto/get_meta.dart
ookami-kb, line 1: Why do you need this? There's already `GetTokensMetaResponseDto` exported from `get_tokens`.

ookami-kb:
@Jaxiii please resolve conflicts here.

packages/espressocash_app/lib/features/tokens/data/token_repository.dart
ookami-kb, line 1: More work is needed on this file:

1. Isolate should only be doing parsing of the downloaded file and saving it to the database (one-by-one, no need to do it in batch, if it breaks nothing bad happens, just not all the tokens will be updated). Everything else should happen in the main isolate.
2. There should be 2 parts of filling the database – from the local file (that we ship with assets) – this should be blocking (aka part of the initialization of the app); another part is getting and parsing the file from the backend.
3. No need to clear the table (on user logout or elsewhere). There are no sensitive information in the tokens, and it can be shared between logs in and out.
4. For running an isolate you can use `compute` instead of the low-level isolates API.

ookami-kb (changes requested):

justinenerio (dismissed):

Merculiar (dismissed):

ookami-kb (approved):

PR #1527, opened 2024-07-22
fix: return data type

### LINKED ISSUE #1460: Return data from Solana Anchor program is now a List, rather than a String
https://github.com/espresso-cash/espresso-cash-public/blob/b920d7976a42a3947371087f28c0c478ce1c123a/packages/solana/lib/src/rpc/dto/return_data.dart#L20

If an anchor program returns data, it now comes back as a list with the base64 encoding, and a 'base64' string, rather than just a String on its own. As a result, this decode throws an exception.

image

It's also possible to return a wide range of types, so I'm not sure of the best approach here.

Locally I've converted this to a dynamic type, however I'm not sure of the precise fix you'd want here. LMK and I'll submit a PR.

Jaxiii:
## Changes

ReturnData data changes from String to dynamic

https://solana-labs.github.io/solana-web3.js/types/TransactionReturnData.html

## Related issues

Fixes #1460

## Checklist

- [ ] Tests added.
- [x] Self-review done.


justinenerio (approved):

Merculiar (approved):

ookami-kb:
Superseded by #1567

PR #1533, opened 2024-07-30
refactor: moneygram pr

Merculiar:
## Changes

Fixes of issues for https://github.com/espresso-cash/espresso-cash-public/pull/1506

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/espressocash_app/lib/features/ramp/services/off_ramp_order_service.dart
```diff
@@ -330,13 +330,12 @@ class OffRampOrderService implements Disposable {
       RampPartner.kado => sl<KadoOffRampOrderWatcher>(),
       RampPartner.scalex => sl<ScalexOffRampOrderWatcher>(),
       RampPartner.coinflow => sl<CoinflowOffRampOrderWatcher>(),
-      RampPartner.moneygram => null,
       RampPartner.rampNetwork ||
       RampPartner.moneygram ||
```
ookami-kb, line 334: Why should it throw on moneygram? Isn't previous PR handling it?

Merculiar: Actually I think, this is does not matter, because this method is never reached if RampPartner is moneygram

ookami-kb: Why?

Merculiar:
```
  @PostConstruct(preResolve: true)
  Future<void> init() async {
    final query = _db.select(_db.offRampOrderRows)
      ..where(
        (tbl) => tbl.status.isNotInValues([
          OffRampOrderStatus.completed,
          OffRampOrderStatus.cancelled,
        ]),
      );

    final orders = await query.get();

    for (final order in orders) {
      if (order.partner == RampPartner.moneygram) {
        continue;
      }

      _subscribe(order.id);
      unawaited(_watch(order.id));
    }
  }
```

Merculiar: Same as in on_ramp_order_service, we use another services (MoneygramOnRampOrderService/MoneygramOffRampOrderService ) to subscribe to moneygram order.

ookami-kb: Ok then, let's have it with throwing then, just leave the comment there. And we'll need to rethink this mechanism in the future, it's confusing.

ookami-kb (approved):

PR #1549, opened 2024-08-17
feat: add stellar recovery service

justinenerio:
## Changes

- adds stellar recovery incase there is pending usdc in stellar account
- add stellar address to intercom profile

## Videos
<details><summary>Recovery</summary>

https://github.com/user-attachments/assets/827ac339-bbc5-4d52-8c1e-8bce72bdb609

</details>

<details><summary>Success</summary>

Simulator Screenshot - iPhone 15 Pro - 2024-09-01 at 20 21 48

</details>

<details><summary>Failure</summary>

Simulator Screenshot - iPhone 15 Pro - 2024-09-01 at 20 13 48

</details>

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/espressocash_app/lib/features/stellar/service/recovery_service.dart
```diff
@@ -0,0 +1,116 @@
...
+  final StellarClient _stellarClient;
+  final EspressoCashClient _ecClient;
+
+  final SharedPreferences _storage;
+
+  bool _hasStellarUsdc = false;
+  bool get hasStellarUsdc => _hasStellarUsdc;
+
+  @PostConstruct(preResolve: true)
+  Future<void> init() async {
```
ookami-kb, line 40: I don't think it should be "blocking" call preventing from accessing the application. Imagine that user has bad internet access, so the app would be initializing pretty long. There should be no long calls in dependency initialization.

Also, I would rather have this checking fully asynchronous, and when the check is done, and user has some money in Stellar account (and they didn't initiate it in this instance of the app), we can show some notification.

ookami-kb (changes requested):

packages/espressocash_app/lib/features/stellar/service/recovery_service.dart
justinenerio: @ookami-kb I've updated the code. Shouldn't be blocking now and will show screen asynchronously

ookami-kb (approved):

PR #1555, opened 2024-08-22
fix: moneygram bridge precision

justinenerio:
## Changes

fixes precision when doing moneygram deposit bridge

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/espressocash_app/lib/features/ramp/partners/moneygram/service/moneygram_on_ramp_service.dart
```diff
@@ -283,9 +283,12 @@ class MoneygramOnRampOrderService implements Disposable {
       );
     }
 
-    final amount = CryptoAmount(
-      value: order.amount,
-      cryptoCurrency: Currency.usdc,
+    final amount =
+        Amount.fiat(value: order.amount, fiatCurrency: Currency.usd).let(
```
ookami-kb, line 287: I don't get it, why does it need to be converted to fiat first?

justinenerio: `order.amount` is int value from fiat amount, and bridge needs precision in USDC

ookami-kb (approved):

PR #1572, opened 2024-09-13
ci: update actions

Killpit:
## Changes

Pipelines were having older and non-standard actions uses, so I decided to upgrade them and changed them with the latest versions so that the pipeline becomes modernized and catch up with the most recent actions versions to support more functionality.

PR is ready for review and self-review had been completed


ookami-kb (approved):
Thank you for the contribution, LGTM.

PR #1598, opened 2024-10-24
feat: update moneygram api endpoints

justinenerio:
## Changes

adds priority fee from quote

## Checklist

- [ ] Tests added.
- [x] Self-review done.


packages/espressocash_api/pubspec.yaml
ookami-kb, line 1: Why do you need changes here?

justinenerio: I've reverted the changes here. Generator was updating more because of higher versions of retrofit

ookami-kb (approved):

PR #1619, opened 2024-11-26
fix: android intercom crash

Merculiar:
## Changes

Fixes intercom webview crash in Android.

## Checklist

- [ ] Tests added.
- [x] Self-review done.

### DIFF
--- a/packages/espressocash_app/android/gradle.properties
+++ b/packages/espressocash_app/android/gradle.properties
@@ -2,6 +2,7 @@ org.gradle.jvmargs=-Xmx1536M
 android.useAndroidX=true
 android.enableJetifier=true
 android.enableR8=true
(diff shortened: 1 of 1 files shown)


Merculiar:
@ookami-kb 
The crash is reproduced only in release mode.
There are several issues related to this bug in flutter_intercom:
[https://github.com/v3rm0n/intercom_flutter/issues/478](https://github.com/v3rm0n/intercom_flutter/issues/478)
[https://github.com/v3rm0n/intercom_flutter/issues/383](https://github.com/v3rm0n/intercom_flutter/issues/383)

It can also be fixed in another way - by adding proguard rules, not sure which solution is better.

ookami-kb (approved):

PR #1682, opened 2025-02-05
chore: Remove `pubspec.lock`

AlexV525:
## Changes

Removes the unnecessary `pubspec.lock` since it only covers the workspace's dependencies. Locks should be kept for Application projects as usual practice.

## Checklist

- [ ] Tests added.
- [x] Self-review done.


ookami-kb:
No, that's intentional. Top-level `*.lock` file is needed to have consistent and reproducible workspace environment.

AlexV525:
The workspace only resolves for Melos so it is useless unless explicit Dart code is using that dependency? Could you reconsider this?

PR #1725, opened 2025-05-05
feat: expose login on ec client

justinenerio:
## Changes

- updates ec client to expose login method
- init client with token

## Related issues

Related to https://github.com/espresso-cash/espresso-cash-meta/issues/14

## Checklist

- [ ] Tests added.
- [x] Self-review done.

### DIFF
--- a/packages/espressocash_client/lib/src/espressocash_client.dart
+++ b/packages/espressocash_client/lib/src/espressocash_client.dart
@@ -18,6 +18,20 @@ const defaultBaseUrl = 'grpc.espressocash.com';
 typedef SignRequest = Future<String> Function(Iterable<int> data);
 
 class EspressoCashClient {
+  factory EspressoCashClient({
+    String? baseUrl,
+    int? port,
+    required SignRequest sign,
+    String? walletAddress,
+    bool secure = true,
+  }) => EspressoCashClient._(
+    baseUrl: baseUrl ?? defaultBaseUrl,
+    port: port,
+    sign: sign,
+    walletAddress: walletAddress,
+    secure: secure,
+  );
+
   factory EspressoCashClient.anonymous({String? baseUrl, int? port, bool secure = true}) =>
       EspressoCashClient._(
         baseUrl: baseUrl ?? defaultBaseUrl,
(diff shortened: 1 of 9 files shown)


packages/espressocash_client/lib/src/espressocash_client.dart
```diff
@@ -122,7 +129,15 @@ class EspressoCashClient {
     return response.token;
   }
 
+  void _ensureLoggedIn() {
```
ookami-kb, line 132: I don't think it's needed. Clients should check for unauthenticated error response and handle it correctly. They have to do it anyway, because having saved token is not a guarantee that it's still valid on the backend.

ookami-kb (changes requested):

packages/espressocash_client/lib/src/espressocash_client.dart
justinenerio: makes sense. refactored and update client

packages/espressocash_client/lib/src/espressocash_client.dart
```diff
@@ -18,6 +18,20 @@ const defaultBaseUrl = 'grpc.espressocash.com';
...
+    bool secure = true,
+  }) => EspressoCashClient._(
+    baseUrl: baseUrl ?? defaultBaseUrl,
+    port: port,
+    sign: sign,
+    walletAddress: walletAddress,
+    secure: secure,
+  );
+
   factory EspressoCashClient.anonymous({String? baseUrl, int? port, bool secure = true}) =>
```
ookami-kb, line 35: Do we need `anonymous` constructor at all?

packages/espressocash_client/lib/src/espressocash_client.dart
```diff
@@ -53,55 +74,44 @@ class EspressoCashClient {
...
+    _moneygramServiceClient = MoneygramServiceClient(_channel, options: callOptions);
+    _tokensServiceClient = TokensServiceClient(_channel, options: callOptions);
   }
 
   Future<void> dispose() async {
     await _channel.shutdown();
   }
 
-  Future<void> _init() async {
+  Future<String> login() async {
```
ookami-kb, line 102: Let's probably make `login` accept optional `token` parameter, and make `initWithToken` private method, to simplify the client's flow.

packages/espressocash_client/lib/src/espressocash_client.dart
justinenerio: in current code it is used here: https://github.com/espresso-cash/espresso-cash-public/blob/6c2b38f05f53d895e733665edd7f8de12118f3a9/packages/espressocash_app/lib/features/dynamic_links/data/dynamic_links_client.dart#L8

packages/espressocash_client/lib/src/espressocash_client.dart
justinenerio: updated login api

ookami-kb (approved):

